PR #78, opened 2026-06-01
feat(insight): event_id on Event-source insights + Event Type filter for all account types

RamanKharchee:
# Description

Two related event-surface improvements:

1. **Insight `event_id` (backend).** Event-source insights (`source: "Event"`, `EventAggregation` type) now carry a representative `event_id` on each entry of the `applications` array, so consumers can link an application back to a concrete event.
2. **Event Type filter (frontend).** The Troubleshoot **Events** sub-tab only showed the **Event Type** dropdown for K8s accounts. It now renders for all account types (K8s + AWS/GCP/Azure), matching the grouped event tables which already show it unconditionally.

No DB migration required — `applications` is already JSON-serialized into the `insight` table, so the new field rides along.

## Type of change

- [x] Enhancement (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] `go build ./insight/...` + `go vet ./insight/...` pass (api-server/services)
- [x] `gofmt` clean on changed Go files
- [x] Frontend: `prettier --check` clean, `oxlint` 0 errors on KubernetesEvents.jsx
- [x] Manual: Troubleshoot → Events sub-tab → select an AWS/GCP/Azure account → Event Type dropdown now appears

---

# Review Notes

## 1. Changes Involved
- Added `EventID string \`json:"event_id,omitempty"\`` to `RelevantApplications`.
- Rewrote the `processEventAggregationRule` SQL: inner subquery `GROUP BY` per application with `(array_agg(id ORDER BY created_at DESC))[1] AS event_id`; outer `json_agg(jsonb_build_object(...))` now emits `event_id`.
- Moved the `filter-event-type` `FilterDropdown` out of the `accountType === 'K8s'` block into an unconditional position (gated only by `!disabledFilters.includes('aggregationKey')`).

## 2. Files & Functions Changed
| File | Function / Section | Change |
|------|-------------------|--------|
| `api-server/services/insight/model.go` | `RelevantApplications` | Added `EventID` field |
| `api-server/services/insight/executor.go` | `processEventAggregationRule()` | Query groups per app + attaches latest event id; emits `event_id` |
| `app/src/components1/events/KubernetesEvents.jsx` | Toolbar filters | Event Type dropdown now renders for all account types |

## 3. Impact Analysis — Other Functions
`RelevantApplications` is shared by `Insight` and `InsightListResponse` and is JSON-serialized on insight write/read — the new field flows through both paths automatically. `TraceAggregation` (also `source: "Event"`) is intentionally untouched: it queries the traces table and has no event id. No DB schema or Hasura change.

## 4. Impact Analysis — Performance
Negligible. The inner `DISTINCT` was replaced by an equivalent `GROUP BY` over the same keys plus a per-group `array_agg` — same row cardinality out, one extra aggregate over rows already scanned.

## 5. Impact Analysis — UX
Cloud (AWS/GCP/Azure) accounts in the Troubleshoot Events sub-tab gain an Event Type filter alongside the existing Event Name / Service Name. K8s behavior is unchanged (one Event Type dropdown, just reordered in the toolbar).

## 6. Risks & Counterarguments
- **Representative-only `event_id`.** When many events share an application within the rule window, only the most recent event's id is surfaced. Accepted as the natural "drill-in" target; revisit to an id array if a consumer needs all events.
- **Cloud `aggregationKeyFilter` population.** The now-always-rendered Event Type dropdown relies on `useKubernetesEventFilters` populating options for cloud accounts. The grouped tables already render it unconditionally for cloud, so options are known to populate; if a cloud account returns no aggregation keys the dropdown simply shows empty (no regression).

### DIFF
--- a/api-server/services/insight/executor.go
+++ b/api-server/services/insight/executor.go
@@ -874,6 +874,10 @@ func (re *ruleExecutor) processEventAggregationRule(rule InsightRule, accountIds
 		rangeFilter = fmt.Sprintf("created_at > now() - interval '%d %s'", rule.Range, rule.RangeUnit)
 	}
 
+	// Inner query collapses to one row per application (subject_name/namespace) and
+	// carries the most recent event id for that application via array_agg(... ORDER BY
+	// created_at DESC)[1]. This keeps event_count == distinct-application count (same
+	// threshold semantics as before) while letting each application carry an event_id.
 	query := fmt.Sprintf(`
 		SELECT
 			sub.%s AS tenant,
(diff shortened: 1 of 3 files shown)


RamanKharchee:
Closing — opened against the wrong repository (nudgebee/nudgebee is an unrelated orphan repo). Re-opened against nudgebee/nudgebee-enterprise.

PR #174, opened 2026-06-15
fix(llm-server): replace unsafe type assertions with safe extractRequestMap helper

### LINKED ISSUE #171: fix(llm-server): unchecked type assertion in 13 chains.go handlers panics on non-map `request` field
## Summary

Thirteen handlers in `llm/llm-server/api/chains.go` extract a nested `"request"` field from an incoming RPC payload, but only check for *non-nil* before type-asserting to `map[string]any`. A client sending `{"request": "foo"}` or `{"request": [1, 2]}` makes the assertion fail at runtime and panic — Gin's recovery middleware converts this to HTTP 500, but the request fails and a panic is logged.

## Where (13 sites)

`llm/llm-server/api/chains.go` lines **180, 567, 950, 1102, 1180, 1273, 1359, 1458, 1733, 1805, 1877, 1967, 2044**.

Every site looks exactly like this:

```go
actionRequestPayload := actionRequest.Input
if actionRequestPayload["request"] != nil {
    actionRequestPayload = actionRequestPayload["request"].(map[string]any)  // panics if not a map
}
```

## Repro

Send any RPC body where `request` is non-nil but not an object:

```bash
curl -X POST http://localhost:9999/<endpoint> \
  -H "Content-Type: application/json" \
  -d '{"input": {"request": "not-a-map"}}'
```

Server logs: `panic: interface conversion: interface {} is string, not map[string]any`. Response: HTTP 500.

## Suggested fix

Switch to the comma-ok form so a non-map value falls through to the existing nil-check that follows:

```go
if req, ok := actionRequestPayload["request"].(map[string]any); ok {
    actionRequestPayload = req
}
```

Because the pattern is verbatim across 13 sites, it's worth extracting a small helper so future handlers can't reintroduce the bug:

```go
// In a shared file (e.g. api/payload.go):
func extractRequestMap(input map[string]any) map[string]any {
    if req, ok := input["request"].(map[string]any); ok {
        return req
    }
    return input
}
```

Then each of the 13 sites becomes:

```go
actionRequestPayload := extractRequestMap(actionRequest.Input)
```

## Why good-first-issue

- Mechanical change repeated 13 times — easy to scope, hard to get wrong.
- Touches a single file (plus optionally one helper file).
- No domain knowledge required — pure Go type-safety hardening.
- Suggested fix shape is already in the issue body.

## Validation

After the change:
- `cd llm/llm-server && make validate` should pass.
- The repro curl above should return a structured error, not a 500 from a panic.

## Notes for the contributor

- Test files near the handlers may need a new case for "request is not a map" → falls back to the outer input.
- Don't change the behavior when `request` is absent or nil — only when it's present but not a map.

onurege3467:
Closes #171

Thirteen handlers in `chains.go` use an unchecked type assertion on the `"request"` field of incoming RPC payloads:

```go
if actionRequestPayload["request"] != nil {
    actionRequestPayload = actionRequestPayload["request"].(map[string]any)  // panics if not a map
}
```

A client sending `{"request": "not-a-map"}` triggers a panic at runtime (Gin recovers to HTTP 500, but the request fails).

### What changed

1. **New `extractRequestMap` helper** in `api/common.go` — uses the comma-ok form so non-map values fall through safely to the original input.

2. **All 13 sites in `chains.go`** now call `extractRequestMap(actionRequest.Input)` instead of the unsafe 3-line pattern.

### Behaviour preserved

- `request` is a map → extracted as before (happy path unchanged)
- `request` is absent or nil → returns original `Input` (falls through to existing nil-check / `requestMap` fallback)
- `request` is present but not a map → returns original `Input` (was a panic, now safe)

### Verification

Platform constraint: no Go toolchain on Windows. The change is a straightforward comma-ok extraction — identical logic to the existing safe pattern in `agents.go` (line 412-420):

```go
payload := actionRequest.Input
if rawRequest, ok := payload["request"]; ok {
    if castedRequest, castOk := rawRequest.(map[string]any); castOk {
        payload = castedRequest
    }
}
```

The project's `make validate` / CI on Linux will confirm full compilation and test suite.

### DIFF
--- a/llm/llm-server/api/chains.go
+++ b/llm/llm-server/api/chains.go
@@ -175,10 +175,7 @@ func handleCompletionApis(r *gin.Engine, tracer trace.Tracer, meter metric.Meter
 			}))
 			return
 		}
-		actionRequestPayload := actionRequest.Input
-		if actionRequestPayload["request"] != nil {
-			actionRequestPayload = actionRequestPayload["request"].(map[string]any)
-		}
+		actionRequestPayload := extractRequestMap(actionRequest.Input)
 		if actionRequestPayload == nil {
 			actionRequestPayload = requestMap
 		}
@@ -562,10 +559,7 @@ func handleCompletionApis(r *gin.Engine, tracer trace.Tracer, meter metric.Meter
 			}))
 			return
 		}
-		actionRequestPayload := actionRequest.Input
-		if actionRequestPayload["request"] != nil {
-			actionRequestPayload = actionRequestPayload["request"].(map[string]any)
-		}
+		actionRequestPayload := extractRequestMap(actionRequest.Input)
 		if actionRequestPayload == nil {
 			actionRequestPayload = requestMap
 		}
@@ -945,10 +939,7 @@ func handleCompletionApis(r *gin.Engine, tracer trace.Tracer, meter metric.Meter
 			}))
 			return
 		}
-		actionRequestPayload := actionRequest.Input
-		if actionRequestPayload["request"] != nil {
-			actionRequestPayload = actionRequestPayload["request"].(map[string]any)
-		}
+		actionRequestPayload := extractRequestMap(actionRequest.Input)
 		if actionRequestPayload == nil {
 			actionRequestPayload = requestMap
 		}
@@ -1097,10 +1088,7 @@ func handleCompletionApis(r *gin.Engine, tracer trace.Tracer, meter metric.Meter
 			}))
 			return
 		}
-		actionRequestPayload := actionRequest.Input
-		if actionRequestPayload["request"] != nil {
-			actionRequestPayload = actionRequestPayload["request"].(map[string]any)
-		}
+		actionRequestPayload := extractRequestMap(actionRequest.Input)
 		if actionRequestPayload == nil {
 			actionRequestPayload = requestMap
 		}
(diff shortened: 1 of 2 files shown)


blue4209211:
@onurege3467  can you please sign CLA ? cant merge without that

https://github.com/nudgebee/nudgebee/pull/174#issuecomment-4709413530

onurege3467:
Signed the cla, can u check it? @blue4209211

blue4209211:
> Signed the cla, can u check it? @blue4209211

@onurege3467

Hey, Thanks, I can see ur CLA sign, but Github bot is not approving it.

**Reason -**
ur current commit has email -  onur@example.com which doesn't matches with ur github email, so bot can't verify it.

Quick fix: re-author with an email that's on your account (you can also use the
  no-reply address from https://github.com/settings/emails) and force-push:

  ```bash
  git config user.email "<verified-email-from-github-settings>"
  git commit --amend --reset-author --no-edit
  git push --force-with-lease
```

After the push, CLA assistant will re-run and clear automatically. No need to  re-sign.

Thanks again.

onurege3467:
Done! I've amended the commit with `onurege3467@users.noreply.github.com` (GitHub no-reply) and force-pushed. CLA assistant should re-run automatically — could you confirm it's cleared now?

blue4209211 (approved):

PR #181, opened 2026-06-18
fix(dns): anchor Route 53 zone matching to label boundaries

### LINKED ISSUE #142: Route 53 zone matching crosses label boundaries
## Summary
Zone membership is label-delimited, but the matcher uses an unanchored `strings.HasSuffix`, so `notexample.com` matches zone `example.com`. The function's own doc comment says it should find the zone the hostname actually belongs to.

## Location
- `api-server/services/knowledge_graph/flow_sources/dns_resolver.go:187`

## Current code
```go
zoneName := strings.TrimSuffix(zones[i].Name, ".")
if strings.HasSuffix(hostname, zoneName) && len(zoneName) > bestMatchLen { // unanchored
    bestMatch = &zones[i]
    bestMatchLen = len(zoneName)
}
```

## Suggested fix
```go
if (hostname == zoneName || strings.HasSuffix(hostname, "."+zoneName)) && len(zoneName) > bestMatchLen {
```

## Acceptance criteria
- [ ] `notexample.com` does **not** match zone `example.com`
- [ ] `foo.example.com` still matches zone `example.com`
- [ ] Tests cover both

---

salmankadaya:
## Description

The `findMatchingZone` and direct zone lookup functions used an unanchored
`strings.HasSuffix(hostname, zoneName)`, causing `notexample.com` to
incorrectly match zone `example.com`. The fix requires either an exact
match or a dot-prefixed suffix (`strings.HasSuffix(hostname, "."+zoneName)`)
so zone membership is always label-delimited.

Fixes #142 

## Changes

- `dns_resolver.go:187` — `findMatchingZone`: anchored best-match loop
- `dns_resolver.go:369` — direct zone lookup: same label-boundary guard

## Acceptance criteria

- [x] `notexample.com` does **not** match zone `example.com`
- [x] `foo.example.com` still matches zone `example.com`
- [x] `example.com` (exact) still matches zone `example.com`

## Type of change

### DIFF
--- a/api-server/services/knowledge_graph/flow_sources/dns_resolver.go
+++ b/api-server/services/knowledge_graph/flow_sources/dns_resolver.go
@@ -184,7 +184,7 @@ func findMatchingZone(hostname string, zones []Route53HostedZone) *Route53Hosted
 
 	for i := range zones {
 		zoneName := strings.TrimSuffix(zones[i].Name, ".")
-		if strings.HasSuffix(hostname, zoneName) && len(zoneName) > bestMatchLen {
+		if (hostname == zoneName || strings.HasSuffix(hostname, "."+zoneName)) && len(zoneName) > bestMatchLen {
 			bestMatch = &zones[i]
 			bestMatchLen = len(zoneName)
 		}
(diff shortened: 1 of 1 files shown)


mayankpande88:
Thanks, the label-boundary fix is correct and covers both call sites.

Gemini raised a valid point: DNS names are case-insensitive and can carry a trailing dot, so something like `foo.Example.com.` would miss the match. Could you lowercase and trim the trailing dot on the hostname before comparing, and add a couple of test cases to cover it?

mayankpande88 (approved):
LGTM. This fixes the label-boundary bug (#142) at both call sites (findMatchingZone and EnrichRoute53DNSWithTargets). The case-insensitivity/trailing-dot handling Gemini raised is a worthwhile robustness improvement but separate from this bug; tracking that separately. CI green.

PR #182, opened 2026-06-18
fix(api-server): use HasPrefix for 172. private-IP check

### LINKED ISSUE #140: Private-IP check uses `Contains` while its three siblings use `HasPrefix`
## Summary
The internal-IP check uses `HasPrefix` for `127.`, `10.`, and `192.168.`, but the `172.` check anomalously uses `strings.Contains`. A public domain like `app-172.prod.example.com` then matches and is misclassified as an internal IP — it becomes a `Service` node instead of an `ExternalService`.

## Location
- `api-server/services/traces/knowledge_graph_service.go:452`

## Current code
```go
if strings.HasPrefix(hostname, "localhost") ||
    strings.HasPrefix(hostname, "127.") ||
    strings.HasPrefix(hostname, "10.") ||
    strings.HasPrefix(hostname, "192.168.") ||
    strings.Contains(hostname, "172.") {   // the odd one out
    return true
}
```

## Suggested fix
```go
strings.HasPrefix(hostname, "172.")
```

## Acceptance criteria
- [ ] `172.` check uses `HasPrefix`
- [ ] Test asserts `app-172.prod.example.com` is treated as external

---

sankur-codes:
# Description
  
  The private-IP detection in `knowledge_graph_service.go` used `strings.Contains` for the `172.` range while the other three checks (`127.`, `10.`, `192.168.`) correctly used `strings.HasPrefix`. Any hostname containing "172." anywhere (e.g. `app-172.prod.example.com`) was misclassified as internal, causing it to render as a `Service` node instead of an `ExternalService` node in the knowledge graph.

Switched to `strings.HasPrefix` and added regression tests.
  
Fixes #140
  
## Type of change
  
# How Has This Been Tested?
  
- [x] Added regression test: `app-172.prod.example.com` is treated as external
- [x] Added positive test: `172.16.0.1` is still treated as internal
- [x] `go test ./traces/ -run TestIsInternalDomain` passes (9/9)

### DIFF
--- a/api-server/services/integrations/servicenow_webhook_enrich_test.go
+++ b/api-server/services/integrations/servicenow_webhook_enrich_test.go
@@ -33,9 +33,9 @@ func TestMergeServiceNowFieldsIntoLabels(t *testing.T) {
 		},
 
 		// Custom u_* string fields
-		"u_cloud_technology":    "AWS",
-		"u_emea_incident_id":    "EMEA-42",
-		"u_environment":         "",
+		"u_cloud_technology":       "AWS",
+		"u_emea_incident_id":       "EMEA-42",
+		"u_environment":            "",
 		"u_custom_vendor_informed": "false",
 
 		// u_payload is a JSON string; top-level scalars should be flattened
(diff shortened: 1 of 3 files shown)


sankur-codes:
@blue4209211 This is the PR for #140 as its suggested there. 

Although, the `HasPrefix` fix resolves the immediate bug, but there's a related correctness gap worth a follow-up: `HasPrefix("172.")` treats all of `172.0.0.0/8` as private, when RFC 1918 only reserves `172.16.0.0/12` (`172.16.*` – `172.31.*`). Public IPs like `172.5.1.1` would still be misclassified as internal.

The codebase already handles this correctly in `knowledge_graph/sources/k8s_source.go` (`isRFC1918IPv4`), which uses `net.ParseIP` and checks `v4[1] >= 16 && v4[1] <= 31`. The traces code could reuse or mirror that approach in a follow-up.

mayankpande88:
This is the right approach. `netip` with `IsPrivate`/`IsLoopback` covers the full 172.16/12 range and correctly rejects public 172.x addresses, which resolves what the bots flagged on the earlier commit.

One CI fix: the `net/netip` import is out of order (it should sort before the `nudgebee` imports), so gofmt will fail. Run `make fmt` in `api-server/services`.

Note #187 targets the same issue; we'll consolidate on this one.

sankur-codes:
> This is the right approach. `netip` with `IsPrivate`/`IsLoopback` covers the full 172.16/12 range and correctly rejects public 172.x addresses, which resolves what the bots flagged on the earlier commit.
> 
> One CI fix: the `net/netip` import is out of order (it should sort before the `nudgebee` imports), so gofmt will fail. Run `make fmt` in `api-server/services`.
> 
> Note #187 targets the same issue; we'll consolidate on this one.

@mayankpande88 Thanks for pointing that out. I've done the needful and also squashed the commit.

mayankpande88:
Correction to my earlier note: CI ran and `Go Lint` (golangci-lint, which includes gofmt) is green, and the committed file actually has `net/netip` ordered correctly. My import-ordering comment was based on a stale diff view, so disregard it. Nothing to change here.

mayankpande88 (approved):
Approving. The netip-based check is correct and CI is green. Keeping this one for #140; #187 can be closed as a duplicate.

PR #187, opened 2026-06-19
fix(api-server): use HasPrefix for 172. private-IP check

### LINKED ISSUE #140: Private-IP check uses `Contains` while its three siblings use `HasPrefix`

shivam2003-dev:
# Description

The internal-IP check in `isInternalDomain` used `strings.Contains` for the `172.` range while its three siblings (`127.`, `10.`, `192.168.`) use `HasPrefix`. A public domain like `app-172.prod.example.com` then matched the substring and was misclassified as an internal `Service` node instead of an `ExternalService`. Switched the `172.` check to `HasPrefix` and added a regression test.

Fixes #140

## Type of change

# How Has This Been Tested?

- [x] `go test ./traces/ -run TestIsInternalDomain` passes
- [x] Added case asserting `app-172.prod.example.com` is treated as external

### DIFF
--- a/api-server/services/traces/knowledge_graph_service.go
+++ b/api-server/services/traces/knowledge_graph_service.go
@@ -449,7 +449,7 @@ func (e *TraceToKnowledgeGraphExtractor) isInternalDomain(hostname string) bool 
 		strings.HasPrefix(hostname, "127.") ||
 		strings.HasPrefix(hostname, "10.") ||
 		strings.HasPrefix(hostname, "192.168.") ||
-		strings.Contains(hostname, "172.") {
+		strings.HasPrefix(hostname, "172.") {
 		return true
 	}
 
--- a/api-server/services/traces/service_classification_test.go
+++ b/api-server/services/traces/service_classification_test.go
@@ -119,6 +119,11 @@ func TestIsInternalDomain(t *testing.T) {
 			description: "Internal IP range",
 		},
 		{
+			hostname:    "172.16.0.1",
+			expected:    true,
+			description: "Internal 172. IP range",
+		},
+		{
 			hostname:    "app.internal",
 			expected:    true,
 			description: "Internal domain suffix",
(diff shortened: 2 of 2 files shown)


mayankpande88:
Thanks for the PR. `HasPrefix` fixes the substring case, but as Gemini, Copilot, and Codex all noted, a `172.` prefix still treats all of 172.0.0.0/8 as private when only 172.16.0.0/12 is, so public 172.x addresses stay misclassified.

PR #182 handles this with `netip.ParseAddr` and `IsPrivate`, which covers the range correctly, so we will likely consolidate on that one. Appreciate the effort, and hope you will pick up other issues.

mayankpande88:
Closing this in favor of #182, which covers issue #140 using `netip.ParseAddr` with `IsPrivate`/`IsLoopback`. That handles the full 172.16/12 range and correctly rejects public 172.x addresses, which the prefix check alone could not. Thanks for the effort here, and please do take a look at other open issues.

PR #188, opened 2026-06-19
fix(api-server): use HasSuffix for TLD check in K8s DNS parser

### LINKED ISSUE #139: K8s DNS detection rejects valid service names that contain a TLD substring
## Summary
`LooksLikeK8sServiceName` is meant to reject hostnames that *end* in an external TLD, but it uses `strings.Contains`, which matches the substring anywhere in the name. Valid `service.namespace`-style K8s names then get rejected and silently fail to enrich.

## Location
- `api-server/services/knowledge_graph/flow_sources/enrichment_k8s_dns_parser.go:85`

## Current code
```go
externalTLDs := []string{".com", ".io", ".org", ".net", ".edu", ".gov", ".co", ".app"}
for _, tld := range externalTLDs {
    if strings.Contains(name, tld) { // should be a suffix check
        return false
    }
}
```

## Why it's wrong
| Input | Rejected by | Should be |
|-------|-------------|-----------|
| `payments.coordinator` | `.co` | accepted |
| `svc.networking` | `.net` | accepted |

## Suggested fix
```go
if strings.HasSuffix(strings.ToLower(name), tld) {
    return false
}
```

## Acceptance criteria
- [ ] Suffix check replaces `Contains`
- [ ] Table-test cases for the inputs above added to the parser's `_test.go`

---

shivam2003-dev:
# Description

`LooksLikeK8sServiceName` used `strings.Contains` to reject external TLDs, which matched the substring **anywhere** in the name. Valid `service.namespace`-style K8s names such as `payments.coordinator` (contains `.co`) and `svc.networking` (contains `.net`) were wrongly rejected and silently failed to enrich. Replaced with a case-insensitive `HasSuffix` check and added table tests.

Fixes #139

## Type of change

# How Has This Been Tested?

- [x] `go test ./knowledge_graph/flow_sources/ -run TestLooksLikeK8sServiceNameTLDSuffix` passes
- [x] Added cases for `payments.coordinator`, `svc.networking` (accepted) and genuine external hostnames (still rejected)

### DIFF
--- a/api-server/services/knowledge_graph/flow_sources/enrichment_k8s_dns_parser.go
+++ b/api-server/services/knowledge_graph/flow_sources/enrichment_k8s_dns_parser.go
@@ -82,7 +82,7 @@ func (p *K8sDNSParser) LooksLikeK8sServiceName(name string) bool {
 	// Skip if it's clearly an external hostname
 	externalTLDs := []string{".com", ".io", ".org", ".net", ".edu", ".gov", ".co", ".app"}
 	for _, tld := range externalTLDs {
-		if strings.Contains(name, tld) {
+		if strings.HasSuffix(strings.ToLower(name), tld) {
 			return false
 		}
 	}
--- a/api-server/services/knowledge_graph/flow_sources/enrichment_k8s_dns_parser_test.go
+++ b/api-server/services/knowledge_graph/flow_sources/enrichment_k8s_dns_parser_test.go
@@ -1,0 +1,34 @@
+package flow_sources
+
+import (
+	"testing"
+)
+
+func TestLooksLikeK8sServiceNameTLDSuffix(t *testing.T) {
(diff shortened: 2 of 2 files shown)


mayankpande88:
Thanks for the PR, and the `ToLower` for case handling is a good addition.

Two things from the bots are still open: `ToLower` runs inside the loop (compute it once), and names with a port, path, or trailing dot bypass the suffix check. PR #183 handles the port/trailing-dot case via `CleanServiceName`, so we will likely consolidate on that, though folding your `ToLower` into it would be ideal. Hope to see you on more issues.

mayankpande88:
Closing this in favor of #183, which fixes issue #139 and also handles ports and trailing dots via `CleanServiceName`. Your `ToLower` idea was good and we noted it on that PR. Thanks for contributing, and hope to see you on more issues.

PR #190, opened 2026-06-19
fix(api-server): wrap retry-exhausted error with %w in agent_warehouse

### LINKED ISSUE #162: Wrap retry-exhausted error with `%w` instead of `%v` in `agent_warehouse.go`
## Summary
When all retries are exhausted, the function formats the underlying `lastErr` with `%v`, flattening it into a string. Using `%w` preserves the error chain so callers (and `errors.Is`/`errors.As`) can inspect the root cause. `lastErr` is non-nil on this path (it is set before every `continue` in the retry loop).

## Location
- `api-server/services/internal/database/agent_warehouse.go:182`

## Current code
```go
return nil, fmt.Errorf("request failed after %d retries: %v", maxRetries, lastErr)
```

## Suggested fix
```go
return nil, fmt.Errorf("request failed after %d retries: %w", maxRetries, lastErr)
```

## Acceptance criteria
- [ ] The final retry error wraps `lastErr` via `%w`
- [ ] `make validate` passes

---
🌱 _Good first issue · difficulty: trivial · area: internal/database. Unassigned — comment to claim._

shivam2003-dev:
# Description

When all retries are exhausted, the function formatted the underlying `lastErr` with `%v`, flattening it into a string. Using `%w` preserves the error chain so callers (and `errors.Is`/`errors.As`) can inspect the root cause. `lastErr` is non-nil on this path (it is set before every `continue` in the retry loop).

Fixes #162

## Type of change

# How Has This Been Tested?

- [x] Error now wraps `lastErr` via `%w`, preserving the chain for `errors.Is`/`errors.As`

### DIFF
--- a/api-server/services/internal/database/agent_warehouse.go
+++ b/api-server/services/internal/database/agent_warehouse.go
@@ -179,7 +179,7 @@ func (s agentWarehouseStmt) executeChronosphereRequestWithRetry(requestData map[
 	}
 
 	// All retries exhausted
-	return nil, fmt.Errorf("request failed after %d retries: %v", maxRetries, lastErr)
+	return nil, fmt.Errorf("request failed after %d retries: %w", maxRetries, lastErr)
 }
 
 const driverName = "agent_warehouse"


mayankpande88 (approved):
Approving. Verified that every non-returning path in the retry loop sets `lastErr` before `continue` and the loop always runs at least once (`attempt <= maxRetries`), so `lastErr` is non-nil on the exhausted path and `%w` is safe. This restores the error chain for `errors.Is`/`errors.As`. CI green. Thanks for the fix.

PR #193, opened 2026-06-19
fix(llm): replace debug print with logging.debug in rag-server config

### LINKED ISSUE #136: Replace debug `print()` with proper logging in rag config
## Summary
A leftover `print()` call writes debug output to stdout instead of going through the logger. It should use the module logger so output respects log configuration/levels.

## Location
- `llm/rag-server/config/__init__.py:10` — `print(f"LOGGING_CONFIG_FILE: {config_file_path}")`

## Suggested fix
Use `logging.debug(...)` (or remove if no longer needed).

## Acceptance criteria
- [ ] No debug `print()` remains in the file
- [ ] Output goes through the logger

---
🌱 _Good first issue · difficulty: trivial · area: rag-server. Unassigned — comment to claim._

shivam2003-dev:
# Description

`setup_logger()` printed the resolved `LOGGING_CONFIG_FILE` path with a stray `print()` left over from debugging. This always writes to stdout regardless of log level. Replaced it with `logging.debug()` so it respects configured log level and handlers.

Fixes #136

## Type of change

# How Has This Been Tested?

- [x] Module already imports `logging`; change is a one-line swap to `logging.debug`

### DIFF
--- a/llm/rag-server/config/__init__.py
+++ b/llm/rag-server/config/__init__.py
@@ -7,7 +7,6 @@ import os
 def setup_logger() -> None:
     # Get the path to the logging config file from the environment variable
     config_file_path = os.getenv("LOGGING_CONFIG_FILE")
-    logging.debug("LOGGING_CONFIG_FILE: %s", config_file_path)
 
     if config_file_path:
         try:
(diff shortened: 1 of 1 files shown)


shivam2003-dev:
Good catch — addressed in the latest commit. The `logging.debug()` ran before `dictConfig`/`load_default_config`, so with the root logger defaulting to WARNING the DEBUG line was silently dropped. Moved it to right after `dictConfig(config)`, where the configured handlers/level are active, so it now logs the resolved path effectively.

mayankpande88 (approved):
LGTM, logging is already imported and CI is green. Thanks.

PR #196, opened 2026-06-19
fix(notifications): use LOG.exception instead of print in transformer

### LINKED ISSUE #109: Replace `print()` with logger in base64 text-extraction helpers
## Summary
`extract_text_from_base64_gz()` and `extract_text_from_base64()` report errors with `print(f"Error extracting text: {e}")`. The module already imports `logging`; these should use `LOG.error(...)` for consistency.

## Locations
- `notifications-server/notifications_server/utils/transformer.py:471`
- `notifications-server/notifications_server/utils/transformer.py:481`

## Suggested fix
Replace both `print(...)` calls with `LOG.error("Error extracting text: %s", e)` (or equivalent).

## Acceptance criteria
- [ ] No `print()` calls remain in the module
- [ ] Errors go through the logger
- [ ] `poetry run pytest` passes

---
🌱 _Good first issue · difficulty: trivial · area: notifications-server. Unassigned — comment to claim._

shivam2003-dev:
# Description

The two base64 extract helpers (`extract_text_from_base64_gz` and `extract_text_from_base64`) logged decode failures with `print()`, which always writes to stdout and ignores the configured logging level/handlers. Switched both to `LOG.error(...)` (the module already defines `LOG = logging.getLogger(__name__)`).

Fixes #109

## Type of change

# How Has This Been Tested?

- [x] Verified both call sites now use `LOG.error` with lazy `%s` formatting

### DIFF
--- a/notifications-server/notifications_server/utils/transformer.py
+++ b/notifications-server/notifications_server/utils/transformer.py
@@ -467,8 +467,8 @@ class Transformer:
             encoded_data = encoded_data[2:-1]
             decoded_data = base64.b64decode(encoded_data)
             return gzip.decompress(decoded_data)
-        except Exception as e:
-            LOG.error("Error extracting text: %s", e)
+        except Exception:
+            LOG.exception("Error extracting text from base64 gzip data")
             return None
 
     @staticmethod
@@ -477,8 +477,8 @@ class Transformer:
             encoded_data = encoded_data[2:-1]
             decoded_data = base64.b64decode(encoded_data)
             return decoded_data
-        except Exception as e:
-            LOG.error("Error extracting text: %s", e)
+        except Exception:
+            LOG.exception("Error extracting text from base64 data")
             return None
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


shivam2003-dev:
Thanks for the review — addressed in the latest commit:

- Both helpers now use `LOG.exception(...)` (full stack trace) instead of `LOG.error("...%s", e)`.
- Distinct messages per helper (`...base64 gzip data` vs `...base64 data`) so failures are distinguishable in logs.

Re: the remaining `print(...)` at lines ~69/72 — those are **doctest examples inside the `apply_length_limit` docstring** (`>>> print(...)`), i.e. documentation, not runtime calls. Issue #109 is about the error-swallowing `print()` in the extract helpers, so I've left the doctests intact.

mayankpande88 (approved):
LGTM. Current code already uses LOG.exception with distinct messages; the remaining print() lines are doctest examples and are intended to stay. CI green.

PR #199, opened 2026-06-19
fix(recommendation): log marshalling error instead of the payload

### LINKED ISSUE #158: Recommendation marshal-error log puts the payload in the `error` field instead of `err`
## Summary
The structured-log `"error"` field is given `r` (the recommendation map being marshalled) instead of the actual error `err`. The real marshalling error is silently dropped from logs. Sibling call sites in the same file (e.g. lines 338, 361, 529, 817) all correctly pass `"error", err`.

## Location
- `api-server/services/recommendation/k8s_recommendation_service.go:188`

## Current code
```go
jsonStr, err := common.MarshalJson(r)
if err != nil {
    ctx.GetLogger().Error("error marshalling recommendation", "error", r)
    return err
}
```

## Suggested fix
```go
jsonStr, err := common.MarshalJson(r)
if err != nil {
    ctx.GetLogger().Error("error marshalling recommendation", "error", err)
    return err
}
```

## Acceptance criteria
- [ ] The `"error"` log field carries `err`, not `r`
- [ ] Matches the pattern used by the other `error marshalling ...` calls in this file
- [ ] `make validate` passes

---
🌱 _Good first issue · difficulty: trivial · area: recommendation. Unassigned — comment to claim._

shivam2003-dev:
# Description

The structured-log `"error"` field was given `r` (the recommendation map being marshalled) instead of the actual error `err`, so the real marshalling failure was silently dropped from logs. Sibling call sites in the same file (lines 338, 361, 529, 817) all correctly pass `"error", err`.

Fixes #158

## Type of change

# How Has This Been Tested?

- [x] Package builds (`go build ./recommendation/`)
- [x] `gofmt` clean
- [x] Matches the pattern of the other `error marshalling` calls in this file

### DIFF
--- a/api-server/services/recommendation/k8s_recommendation_service.go
+++ b/api-server/services/recommendation/k8s_recommendation_service.go
@@ -185,7 +185,7 @@ func processAbandonedRecommendations(ctx *security.RequestContext, accountId str
 		}
 		jsonStr, err := common.MarshalJson(r)
 		if err != nil {
-			ctx.GetLogger().Error("error marshalling recommendation", "error", r)
+			ctx.GetLogger().Error("error marshalling recommendation", "error", err)
 			return err
 		}
 


mayankpande88 (approved):
LGTM, CI green. Thanks.

PR #202, opened 2026-06-19
fix(llm-server): kg_get_node header reads name/namespace/cluster from properties

### LINKED ISSUE #143: `kg_get_node` header always prints "(unnamed)" and never shows namespace/cluster
## Summary
`writeKGNodeHeader` reads top-level keys (`name`, `node_type`) from the node map, and `kgNodeMetaPairs` expects top-level `namespace`/`cluster` — but the `KgNode` payload keeps `name`, `namespace`, and `cluster` inside `properties`. So the header always renders `(unnamed)` and the Namespace/Cluster metadata lines never appear. (The data isn't lost — it still shows under the Properties section; the header/metadata block is just broken.)

## Location
- `llm/llm-server/tools/common_kg.go:491`

## Current code
```go
func writeKGNodeHeader(b *strings.Builder, node map[string]any, id string) {
    name := stringFromAny(node["name"]) // KgNode has no top-level "name"
    if name == "" {
        name = "(unnamed)"
    }
    ...
}
```

## Suggested fix
Fall back to `properties["name"]` (and `properties["namespace"]` / `properties["cluster"]`) when the top-level lookup is empty.

## Acceptance criteria
- [ ] Header shows the node name when `properties.name` is set
- [ ] Namespace/Cluster lines render when present
- [ ] Snapshot test on `formatKGGetNodeResponse`

---

shivam2003-dev:
# Description

`writeKGNodeHeader` reads top-level `name`/`node_type`, and `writeKGNodeMetadata` (via `kgNodeMetaPairs`) expects top-level `namespace`/`cluster`. But the `KgNode` payload keeps `name`, `namespace`, and `cluster` inside `properties`, so the header always rendered `(unnamed)` and the Namespace/Cluster metadata lines never appeared (the data was still visible under the Properties section).

Added a small `kgNodeField()` helper that falls back to `properties[key]` when the top-level lookup is empty, and used it in both the header and metadata writers. Added a unit test covering a node whose `name`/`namespace`/`cluster` live under `properties`.

Fixes #143

## Type of change

# How Has This Been Tested?

- [x] `go test ./tools/ -run TestFormatKGGetNodeResponse` passes (incl. new nested-properties case)
- [x] `gofmt` clean
- [x] Existing top-level-keys tests still pass (fallback is backward compatible)

### DIFF
--- a/llm/llm-server/tools/common_kg.go
+++ b/llm/llm-server/tools/common_kg.go
@@ -489,11 +489,26 @@ func formatKGGetNodeResponse(node map[string]any) string {
 }
 
 func writeKGNodeHeader(b *strings.Builder, node map[string]any, id string) {
-	name := stringFromAny(node["name"])
+	name := kgNodeField(node, "name")
 	if name == "" {
 		name = "(unnamed)"
 	}
-	fmt.Fprintf(b, "**%s** (%s) — id: %s\n", name, stringFromAny(node["node_type"]), id)
+	fmt.Fprintf(b, "**%s** (%s) — id: %s\n", name, kgNodeField(node, "node_type"), id)
+}
+
+// kgNodeField looks up a field on the node, falling back to the node's
+// "properties" map when the top-level key is absent or empty. The KgNode
+// payload nests name/namespace/cluster (and other fields) under "properties",
+// so a top-level-only lookup would render "(unnamed)" and drop the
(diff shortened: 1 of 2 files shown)


mayankpande88 (approved):
LGTM, CI green. The nil-map guard is optional (nil-map indexing is panic-safe and returns empty here), so merging as-is. Thanks.

PR #203, opened 2026-06-19
fix(api-server): enforce label boundary in Route 53 zone matching

### LINKED ISSUE #142: Route 53 zone matching crosses label boundaries

sankur-codes:
`findMatchingZone` used an unanchored `strings.HasSuffix` to match hostnames against Route 53 zones, so `notexample.com` incorrectly
  
Also incorporated DNS normalization: both hostname and zone name are lowercased and trailing dots are trimmed before comparison, since DNS names are case-insensitive and may optionally end with a trailing dot..

Fixes #142

## Type of change

# How Has This Been Tested?

- [x] `notexample.com` does not match zone `example.com` (label boundary)
- [x] `foo.example.com` matches zone `example.com` (subdomain)
- [x] `www.sub.example.com` prefers zone `sub.example.com` (most specific)
- [x] `Foo.Example.COM` matches zone `example.com` (case-insensitive)
- [x] `foo.example.com.` matches zone `example.com.` (trailing dot)
- [x] `go test ./knowledge_graph/flow_sources/ -run TestFindMatchingZone` passes (7/7)
- [x] `make fmt` passes

### DIFF
--- a/api-server/services/knowledge_graph/flow_sources/dns_resolver.go
+++ b/api-server/services/knowledge_graph/flow_sources/dns_resolver.go
@@ -182,9 +182,11 @@ func findMatchingZone(hostname string, zones []Route53HostedZone) *Route53Hosted
 	var bestMatch *Route53HostedZone
 	var bestMatchLen int
 
+	hostname = strings.ToLower(strings.TrimSuffix(hostname, "."))
+
 	for i := range zones {
-		zoneName := strings.TrimSuffix(zones[i].Name, ".")
-		if strings.HasSuffix(hostname, zoneName) && len(zoneName) > bestMatchLen {
+		zoneName := strings.ToLower(strings.TrimSuffix(zones[i].Name, "."))
+		if (hostname == zoneName || strings.HasSuffix(hostname, "."+zoneName)) && len(zoneName) > bestMatchLen {
 			bestMatch = &zones[i]
 			bestMatchLen = len(zoneName)
 		}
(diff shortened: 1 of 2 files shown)


sankur-codes:
@blue4209211 @mayankpande88 I saw that the earlier PR #181  is not being responded on so I raised this one with the fix and suggested changes. Please have a look and review/merge accordingly.

mayankpande88:
Thanks for this, and especially for the thorough test file and the case-insensitive/trailing-dot handling. We landed #181 for #142, which fixes the label-boundary bug at both call sites (findMatchingZone and EnrichRoute53DNSWithTargets); #203 only covered findMatchingZone, leaving EnrichRoute53DNSWithTargets unanchored, so we went with the one that covered both. Closing this as the bug is now fixed. The case-insensitivity + trailing-dot normalization you added is a genuine improvement though - if you'd like, a focused follow-up applying ToLower + trailing-dot trim (with your tests) to both call sites would be very welcome.

Issue #116, opened 2026-06-14
Bug: invalid SMTP port is accepted as valid


blue4209211:
## Summary
`_is_valid_smtp_params()` checks the port with `is_valid_port(port)`, but on failure it only logs — it does **not** `return False`. Execution falls through to the final `return True`, so a config with an out-of-range or non-numeric port is treated as valid. The send is then attempted and fails later with a confusing error.

## Location
- `notifications-server/notifications_server/emailer/sender.py:59-61` — the `if not is_valid_port(port):` block logs but doesn't return.

## Suggested fix
Return `False` in the invalid-port branch (e.g. make the final line `return is_valid_port(port)`). While here, switch the log call to `LOG.error` (see the related `LOG.exception` cleanup issue).

## Acceptance criteria
- [ ] An invalid port causes `_is_valid_smtp_params` to return `False`
- [ ] Valid ports still return `True`
- [ ] A unit test covers both

---
🌱 _Good first issue · difficulty: easy · area: notifications-server. Unassigned — comment to claim._

Issue #110, opened 2026-06-14
Use `LOG.error()` instead of `LOG.exception()` for non-exception SMTP validation failures


blue4209211:
## Summary
In the emailer, validation-failure branches call `LOG.exception(...)`, which logs a full stack trace even when no exception is active. These should use `LOG.error()` (or `LOG.warning()`) so logs aren't cluttered with misleading tracebacks.

## Locations
- `notifications-server/notifications_server/emailer/sender.py:54`
- `notifications-server/notifications_server/emailer/sender.py:57`
- `notifications-server/notifications_server/emailer/sender.py:60`

(All inside `_is_valid_smtp_params()`.)

## Suggested fix
Use `LOG.error()` for these validation failures. Reserve `LOG.exception()` for `except` blocks where an exception is actually being handled.

## Acceptance criteria
- [ ] `LOG.exception()` is only used inside `except` blocks
- [ ] Validation failures log via `LOG.error()`/`LOG.warning()`

---
🌱 _Good first issue · difficulty: trivial · area: notifications-server. Unassigned — comment to claim._

shivam2003-dev:
I'm working on this — opened a PR with the fix. Please assign it to me.

reckless-sherixx:
I have already raise a pr kindly review it #207

PR #207, opened 2026-06-19
fix(notifications): use LOG.error for validation failures

reckless-sherixx:
## Description

This PR fixes an issue in the emailer where validation-failure branches were incorrectly using `LOG.exception(...)`. Since `LOG.exception` automatically logs the current exception stack trace, this was resulting in misleading tracebacks cluttering the logs when no exception was actually active. 

This replaces the `LOG.exception()` calls with `LOG.error()` during SMTP parameter validation.

**Changes made:**
- Updated lines 54, 57, and 60 in `notifications-server/notifications_server/emailer/sender.py` (inside `_is_valid_smtp_params()`) to use `LOG.error()`.
- Reserved `LOG.exception()` strictly for `except` blocks where a genuine exception is being handled.

## Acceptance Criteria
- [x] `LOG.exception()` is only used inside `except` blocks.
- [x] Validation failures log via `LOG.error()`.

## Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/notifications-server/notifications_server/emailer/sender.py
+++ b/notifications-server/notifications_server/emailer/sender.py
@@ -58,6 +58,7 @@ def _is_valid_smtp_params(params):
         return False
     if not is_valid_port(port):
         LOG.error("port %s is not valid ", port)
+        return False
     return True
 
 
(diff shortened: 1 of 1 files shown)


mayankpande88:
Thanks for this, and good catch on the missing `return False` for the invalid-port case. We've split this across the two underlying issues: the missing return landed via #217 (Fixes #116), and the LOG.exception->LOG.error change is being handled in #210 (Fixes #110). Since this PR bundled both without an issue link and now conflicts with main, closing it in favor of those two. Appreciated, and hope you'll pick up other issues.

PR #213, opened 2026-06-19
Fix email regex anchor

antono4:
## Summary

Replace inline email regex with shared `EmailRegEx` from `@lib/validation`.

## Problem

The inline pattern was missing the leading `^` anchor:
```ts
// BUGGY - accepts "random text a@b.com"
const emailPattern = new RegExp('[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+.[a-zA-Z]{2,4}$');


mayankpande88:
Closing — the email-regex fix shipped via #219 (merged), and the other changes on this stacked branch (firing status, memorylimit, operator precedence) shipped via #235/#236/#211. Thanks for the contributions; they were stacked into one branch so we landed the focused, CLA-signed versions.

PR #215, opened 2026-06-19
Fix memorylimit typeerror

antono4:
## Summary

Fix potential `TypeError` when `memoryLimit` contains no digits.

## Problem

```jsx
// BUGGY - throws TypeError if match returns null
setMemoryLimit(parseInt(memoryLimit.match(/\d+/)[0], 10));


antono4:
const memMatch = memoryLimit.match(/\d+/);
if (memMatch) {
  setMemoryLimit(parseInt(memMatch[0], 10));
}

mayankpande88:
Closing in favor of #236 (merged), which fixes the memoryLimit parse with a null guard. Thanks.

PR #219, opened 2026-06-19
fix(ui): use shared EmailRegEx for signin validation

### LINKED ISSUE #149: Sign-in email regex is missing the leading `^` anchor (false-accepts junk before the address)
## Summary
The email-validation pattern in the sign-in page has a trailing `$` but **no leading `^` anchor**, so `RegExp.test()` returns true for any string that merely *ends* with something email-shaped — e.g. `"random text a@b.co"` passes. The dot before the TLD is also unescaped. The repo already exports a correct, anchored `EmailRegEx` in `lib/validation.ts`.

## Location
- `app/src/pages/signin_v2.tsx:245`

## Current code
```ts
const emailPattern = new RegExp('[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+.[a-zA-Z]{2,4}$');
```

## Suggested fix
Reuse the shared, anchored validator instead of a hand-rolled pattern:
```ts
import { EmailRegEx } from '@lib/validation';
// ...
const emailPattern = EmailRegEx;
```

## Acceptance criteria
- [ ] `"foo bar a@b.com"` is rejected as invalid
- [ ] A normal valid email still passes
- [ ] The page reuses the shared `EmailRegEx` rather than an inline pattern

---
🌱 _Good first issue · difficulty: easy · area: ui/auth. Unassigned — comment to claim._

reckless-sherixx:
Replaced inline unanchored regex with the shared EmailRegEx in sign-in page. Fixes #149

### DIFF
--- a/app/src/pages/signin_v2.tsx
+++ b/app/src/pages/signin_v2.tsx
@@ -242,7 +242,7 @@ export default function SignInV2({ providers, samlEnabled, tier }: any) {
       credsPassword: '',
     }));
 
-    const emailPattern = new RegExp('[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+.[a-zA-Z]{2,4}$');
+    const emailPattern = EmailRegEx;
     let hasError = false;
 
     if (!emailPattern.test(credsUsername)) {


mayankpande88 (approved):
LGTM. Uses the shared EmailRegEx (already imported and used elsewhere in this file), CI green. Thanks.

PR #221, opened 2026-06-19
Fix kg utc time

antono4:
## Summary

Use UTC time in knowledge-graph build window.

## Problem

`time.Now()` returns local time, which can cause issues on non-UTC hosts.

## Fix

```go
// Before
now := time.Now()

// After
now := time.Now().UTC()


mayankpande88:
Closing in favor of #237 (merged), which makes the KG build window UTC. Thanks.

PR #225, opened 2026-06-19
fix(ui): prevent NaN in k8s node progress bars

### LINKED ISSUE #150: Node ProgressBar shows `NaN` instead of `0` (`.toFixed(2) || 0` fallback is dead)
## Summary
The `|| 0` fallback here is dead code: `Number(...).toFixed(2)` always returns a non-empty (truthy) string, so `|| 0` can never fire. When `r.values[0]` is missing, `Number(undefined)` is `NaN`, and `NaN.toFixed(2)` is the truthy string `"NaN"` — which is then passed to `ProgressBar` as its value instead of the intended `0`.

## Location
- `app/src/components1/k8s/details/KubernetesNodes.jsx:434`

## Current code
```jsx
value={Number(r.values[0]).toFixed(2) || 0}
```

## Suggested fix
```jsx
value={(Number(r.values[0]) || 0).toFixed(2)}
```

## Acceptance criteria
- [ ] When `r.values[0]` is undefined/non-numeric, the bar shows `0` (e.g. `"0.00"`), not `"NaN"`
- [ ] Valid numeric values still render to 2 decimals

---
🌱 _Good first issue · difficulty: trivial · area: ui/k8s-nodes. Unassigned — comment to claim._

reckless-sherixx:
## Description

This PR fixes a bug in the `KubernetesNodes` component where progress bars would incorrectly render `NaN` instead of `0` when a node metric value was missing or non-numeric.

The issue occurred because the `|| 0` fallback was placed *after* the `.toFixed(2)` method call. Since `.toFixed(2)` on `NaN` returns the truthy string `"NaN"`, the fallback never triggered.

**Changes made:**
- Wrapped the value cast inside the parenthesis so the `|| 0` fallback is evaluated *before* `.toFixed(2)` is called: `(Number(r.values[0]) || 0).toFixed(2)`.
- Applied this fix to all three affected progress bars in `app/src/components1/k8s/details/KubernetesNodes.jsx` (CPU usage, Memory usage, and Requested Memory).

Fixes #150

## Type of change

- [ ] Enhancement (non-breaking change which adds functionality)
- [ ] Refactor (non-breaking change which improves code structure)

# How Has This Been Tested?

- [x] Verified that progress bars properly show "0.00" when the node metric value is undefined or missing.
- [x] Verified that valid numeric values still correctly render to 2 decimal places.

### DIFF
--- a/app/src/components1/k8s/details/KubernetesNodes.jsx
+++ b/app/src/components1/k8s/details/KubernetesNodes.jsx
@@ -431,7 +431,7 @@ const KubernetesNodesTable = ({ accountId, heading = 'All Nodes' }) => {
                         <ProgressBar
                           blueVarient={true}
                           capacity={dataItem[0].drilldownQuery?.cpu_capacity ? `${dataItem[0].drilldownQuery?.cpu_capacity}` : 0}
-                          value={Number(r.values[0]).toFixed(2) || 0}
+                          value={(Number(r.values[0]) || 0).toFixed(2)}
                           largeVariant={true}
                           tooltipRequired={true}
                           showCapacity={false}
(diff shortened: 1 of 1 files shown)


mayankpande88:
Thanks - the fix itself looks correct (wrapping in Number() with a || 0 guard, and good catch covering the third progress bar). CI is red only on Prettier formatting for KubernetesNodes.jsx though. Could you run `npm run prettier:fix` (or `npx prettier --write app/src/components1/k8s/details/KubernetesNodes.jsx`) and push? Should go green after that.

reckless-sherixx:
Updating and and pushing again @mayankpande88

mayankpande88 (approved):
LGTM, Prettier is green now. Thanks.

PR #226, opened 2026-06-19
fix(kg): make protocol inference deterministic (#144)

### LINKED ISSUE #144: Nondeterministic protocol inference due to map iteration
## Summary
`inferProtocol` ranges a `map[string]string` and returns the first `Contains` match. Go map iteration order is randomized, so an operation string matching two patterns can yield different protocols across runs.

## Location
- `api-server/services/knowledge_graph/flow_sources/datadog_apm_flow_source.go:498`

## Current code
```go
for pattern, protocol := range protocolMap {
    if strings.Contains(operation, pattern) {
        return protocol
    }
}
```

## Suggested fix
Replace the map with an ordered slice iterated most-specific-first (deterministic matching).

## Acceptance criteria
- [ ] Result is stable across repeated calls for the same input
- [ ] Test calls the function repeatedly and asserts a stable result

---

shivam2003-dev:
## Summary
`inferProtocol` ranged over a `map[string]string` and returned the first `Contains` match. Go map iteration order is randomized, so an operation string matching two patterns could yield different protocols across runs. Replaced the map with an ordered slice iterated most-specific-first.

## Changes
- `datadog_apm_flow_source.go`: ordered `protocolPatterns` slice instead of a map.
- `datadog_apm_flow_source_test.go`: add `TestDatadogAPMFlowSource_inferProtocol_Deterministic` that calls the function repeatedly for multi-pattern operations and asserts a stable result.

## Acceptance criteria
- [x] Result is stable across repeated calls for the same input
- [x] Test calls the function repeatedly and asserts a stable result

Fixes #144

### DIFF
--- a/api-server/services/knowledge_graph/flow_sources/datadog_apm_flow_source.go
+++ b/api-server/services/knowledge_graph/flow_sources/datadog_apm_flow_source.go
@@ -481,23 +481,28 @@ func (s *DatadogAPMFlowSource) buildEdgeProperties(
 func (s *DatadogAPMFlowSource) inferProtocol(operation string) string {
 	operation = strings.ToLower(operation)
 
-	protocolMap := map[string]string{
-		"http.request":   "http",
-		"grpc":           "grpc",
-		"postgres.query": "postgres",
-		"redis.command":  "redis",
-		"kafka.consume":  "kafka",
-		"kafka.produce":  "kafka",
-		"mongo.query":    "mongo",
-		"s3.command":     "s3",
-		"universal.http": "http",
-		"web.request":    "http",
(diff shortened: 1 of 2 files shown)


mayankpande88 (approved):
LGTM. Ordered slice removes the map-iteration nondeterminism and the added test locks it in. CI green. Thanks.

mayankpande88 (approved):
LGTM. Fixes the nondeterminism (#144) and the test locks it in. Re the per-call allocation Gemini noted: the original map literal was also allocated per call, so this is not a regression - hoisting the table to a package-level var would be a nice follow-up optimization but isn't blocking. Merging.

PR #230, opened 2026-06-19
fix(ticket-server): consistent default/cap for List limit across providers (#129)

### LINKED ISSUE #129: Inconsistent default/cap for `limit` across `List` providers
## Summary
GitHub's `List` defaults `limit` to 25 when `<= 0` and caps it at 100. ZenDuty has no default: with `limit = 0` the client-side slice becomes `incidents[offset:offset]`, returning an **empty page**. A caller that omits `limit` gets wildly different results depending on the provider.

## Locations
- `ticket-server/services/tools/github_service.go:524-529` — reference: default 25 / cap 100
- `ticket-server/services/tools/zenduty_service.go:183` — `end := params.Offset + params.Limit` (empty page when `Limit == 0`)

## Suggested fix
Introduce a shared default + cap (small helper or constants) and apply it consistently in every provider's `List` when `limit <= 0` or exceeds the cap.

## Acceptance criteria
- [ ] An unset/zero `limit` yields the same sensible default page size across providers
- [ ] The cap is applied consistently
- [ ] Behavior is documented

---
🌱 _Good first issue · difficulty: medium · area: ticket-server. Unassigned — comment to claim._

shivam2003-dev:
## Summary
Providers handled an unset `limit` inconsistently. GitHub/GitLab defaulted to 25 and capped at 100, while ZenDuty had no default — `end := params.Offset + params.Limit` made the slice `incidents[offset:offset]`, returning an **empty page** when `limit == 0`.

## Changes
- New `services/tools/pagination.go` with shared constants `DefaultListLimit = 25`, `MaxListLimit = 100` and helpers `normalizeLimit` / `normalizeOffset`.
- `github_service.go` and `gitlab_service.go`: replaced the duplicated inline default/cap with the shared helpers.
- `zenduty_service.go`: normalize limit/offset before client-side slicing, fixing the empty-page bug.

## Acceptance criteria
- [x] An unset/zero `limit` yields the same default page size (25) across providers
- [x] The cap (100) is applied consistently
- [x] Behavior is documented (godoc on the constants/helpers)

## Validation
`gofmt -l` clean; `go build ./services/tools/` passes.

Fixes #129

### DIFF
--- a/ticket-server/services/tools/github_service.go
+++ b/ticket-server/services/tools/github_service.go
@@ -521,16 +521,8 @@ func (s *GitHubService) List(ctx *gin.Context, config models.TicketConfiguration
 	// Convert offset/limit to page/perPage using local variables to avoid
 	// mutating the caller's struct. GitHub caps PerPage at 100 server-side,
 	// so we cap here too so the page calculation matches what is actually returned.
-	limit := params.Limit
-	if limit <= 0 {
-		limit = 25
-	} else if limit > 100 {
-		limit = 100
-	}
-	offset := params.Offset
-	if offset < 0 {
-		offset = 0
-	}
+	limit := normalizeLimit(params.Limit)
+	offset := normalizeOffset(params.Offset)
 	page := (offset / limit) + 1
 
 	opts := &github.IssueListByRepoOptions{
--- a/ticket-server/services/tools/github_service.go
+++ b/ticket-server/services/tools/github_service.go
@@ -518,11 +518,14 @@ func (s *GitHubService) List(ctx *gin.Context, config models.TicketConfiguration
 	}
 	owner, repo := parts[0], parts[1]
 
-	// Convert offset/limit to page/perPage using local variables to avoid
-	// mutating the caller's struct. GitHub caps PerPage at 100 server-side,
-	// so we cap here too so the page calculation matches what is actually returned.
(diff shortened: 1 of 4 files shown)


mayankpande88:
Thanks - the shared normalizeLimit/normalizeOffset helpers are exactly the right call for #129. One thing to tighten while you're here, per Gemini's review: since params is passed by value, assign the normalized values back to params.Limit/params.Offset at the top rather than into separate locals. Otherwise the total calculation (resp.LastPage * params.Limit) and the returned ListResult still carry the un-normalized values (e.g. Limit: 0), which is inconsistent with the page size actually used. A two-line change (params.Limit = normalizeLimit(params.Limit); params.Offset = normalizeOffset(params.Offset)) makes the whole function consistent.

shivam2003-dev:
Addressed the review: limit/offset are now normalized in-place on `params` (it's passed by value, so this is safe), so the GitHub total estimate and every provider's returned `ListResult` reflect the normalized page size instead of an unset 0. ZenDuty now returns `incidents[:0]` rather than nil when offset exceeds total. gofmt clean, build passes.

mayankpande88 (approved):
LGTM. Normalizing params.Limit/Offset in-place fixes the total and ListResult consistency (Gemini's point), and the shared helpers cover all providers (Fixes #129). CI green.

PR #236, opened 2026-06-19
fix(ui): guard memoryLimit digit parse in node pool form

### LINKED ISSUE #151: `memoryLimit.match(/\d+/)[0]` can throw `TypeError` when the value has no digits
## Summary
The `if (memoryLimit)` guard only checks that the string is non-empty. If `memoryLimit` is a non-empty string with **no digits**, `String.match(/\d+/)` returns `null` and the `[0]` access throws a `TypeError`, breaking the node-pool edit form load.

## Location
- `app/src/components1/k8s/details/KubernetesAutoScalerNodePool.jsx:643`

## Current code
```jsx
if (memoryLimit) {
  setMemoryLimit(parseInt(memoryLimit.match(/\d+/)[0], 10));
}
```

## Suggested fix
```jsx
const memMatch = memoryLimit.match(/\d+/);
if (memMatch) {
  setMemoryLimit(parseInt(memMatch[0], 10));
}
```

## Acceptance criteria
- [ ] A `memoryLimit` with no digits no longer throws
- [ ] A normal value like `"32Gi"` still parses to `32`

---
🌱 _Good first issue · difficulty: easy · area: ui/k8s-autoscaler. Unassigned — comment to claim._

Ramneek-kalra:
# Description

The node-pool edit form called `memoryLimit.match(/\d+/)[0]` without checking for a null match. A non-empty value with no digits (e.g. `"Gi"`) threw a `TypeError` and broke form load.

Fixes #151

## Type of change

# How Has This Been Tested?

- [x] Guard only parses when `match(/\d+/)` returns a match
- [ ] `npm run lint2` in `app/` (pending local run)

### DIFF
--- a/app/src/components1/k8s/details/KubernetesAutoScalerNodePool.jsx
+++ b/app/src/components1/k8s/details/KubernetesAutoScalerNodePool.jsx
@@ -639,8 +639,9 @@ const KubernetesAutoScalerNodePool = ({ accountId }) => {
       setInstanceZone(zoneRequirement);
       setCPULimit(cpuLimit);
       setNodeClassRef(nodeClassRefName);
-      if (memoryLimit) {
-        setMemoryLimit(parseInt(memoryLimit.match(/\d+/)[0], 10));
+      const memMatch = memoryLimit?.match(/\d+/);
+      if (memMatch) {
+        setMemoryLimit(parseInt(memMatch[0], 10));
       }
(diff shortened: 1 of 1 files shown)


mayankpande88 (approved):
LGTM. Null-checks the regex match before indexing, preventing the NaN/throw. CI green. The remaining edge-case note is non-blocking.

PR #179, opened 2026-06-18
fix(kg): handle trailing-slash URLs in extractGCPResourceNameFromURL

jenistenxavier:
# Description

`extractGCPResourceNameFromURL` returned `""` for GCP self-link URLs ending in `/` (e.g. `.../networks/default/`), producing an empty lookup key and a dropped knowledge-graph edge. This mirrors the sibling helper `extractGCPShortName`: when the last `/`-split segment is empty, fall back to the second-to-last segment.

Fixes #141

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] Added `{"trailing slash", ".../networks/default/", "default"}` case to `TestExtractGCPResourceNameFromURL`
- [x] `go test ./knowledge_graph/sources/ -run TestExtractGCPResourceNameFromURL` passes (5/5)
- [x] `go vet ./knowledge_graph/sources/` clean

PR #238, opened 2026-06-19
fix(api-server): handle trailing-slash GCP resource URLs

### LINKED ISSUE #141: `extractGCPResourceNameFromURL` returns `""` for a trailing-slash URL
## Summary
`extractGCPResourceNameFromURL` returns the last `/`-split segment, which is `""` when the URL ends in `/` (e.g. `.../networks/default/` yields `""` instead of `default`), producing an empty lookup key and a dropped edge. The sibling helper `extractGCPShortName` (~30 lines above) already handles this case correctly.

## Location
- `api-server/services/knowledge_graph/sources/gcp_source.go:3076-3087`

## Current code
```go
parts := strings.Split(url, "/")
if len(parts) > 0 {
    return parts[len(parts)-1] // "" when the URL ends in "/"
}
return url
```

## Suggested fix
Mirror `extractGCPShortName`: if the last segment is empty and `len(parts) >= 2`, return `parts[len(parts)-2]`.

## Acceptance criteria
- [ ] Trailing-slash URLs return the resource name
- [ ] Add `{"trailing slash", ".../networks/default/", "default"}` to the function's test

---

Ramneek-kalra:
# Description

`extractGCPResourceNameFromURL` returned an empty string for URLs ending in `/` because the last split segment is empty. Mirrored the existing `extractGCPShortName` trailing-slash handling and added a regression test.

Fixes #141

## Type of change

# How Has This Been Tested?

- [x] Added `trailing slash` test case to `TestExtractGCPResourceNameFromURL`
- [ ] `make validate` in `api-server/services` (pending local run)

### DIFF
--- a/api-server/services/knowledge_graph/sources/gcp_source.go
+++ b/api-server/services/knowledge_graph/sources/gcp_source.go
@@ -3079,8 +3079,14 @@ func extractGCPResourceNameFromURL(url string) string {
 	}
 
 	parts := strings.Split(url, "/")
-	if len(parts) > 0 {
-		return parts[len(parts)-1]
+	lastPart := parts[len(parts)-1]
+	if lastPart != "" {
+		return lastPart
+	}
+
+	// If the last part is empty (trailing slash), try the second to last
+	if len(parts) >= 2 {
+		return parts[len(parts)-2]
 	}
 
 	return url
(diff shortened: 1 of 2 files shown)


mayankpande88:
Closing as a duplicate of #179, which fixes the same issue (#141) with an identical change and merged first. Thanks for the fix - both were correct; we consolidated on the earlier one. Your other PRs are being reviewed.

PR #266, opened 2026-06-20
refactor(integrations): extract reusable Grafana alert-to-event mapping (#257)

### LINKED ISSUE #257: refactor(integrations): extract reusable Grafana alert-to-event mapping
## What

Extract the logic that turns a **single Grafana alert into a NudgeBee event** into its own reusable function. This is a pure refactor — no behavior change.

## Why

In `api-server/services/integrations/grafana_webhook.go`, `ProcessEventWebook` does two things in one loop: parses the inbound webhook payload, and maps each alert to an event (fingerprint, firing/resolved status, start/end times, labels, severity, title).

We're about to add a second caller — a job that pulls alerts from Grafana's API and reconciles them — and it needs the **exact same mapping** so both produce identical, `fingerprint`-keyed events (otherwise the two paths create duplicates instead of reconciling). Pulling the per-alert mapping into a shared function lets both reuse it.

## What to do

- Extract the per-alert mapping (the body that builds one `core.EventIncomingWebhook` from one parsed alert) into a standalone function, e.g. `mapGrafanaAlertToEvent(alert map[string]any) (core.EventIncomingWebhook, error)`.
- Have `ProcessEventWebook` call it inside its loop.
- Keep behavior identical.

## Done when

- [ ] One standalone function is the single place that maps a Grafana alert → event
- [ ] `ProcessEventWebook` uses it and its output is unchanged
- [ ] Existing tests pass; add a small unit test for the new function if there isn't one

shivam2003-dev:
## What

Extracts the per-alert Grafana mapping out of `ProcessEventWebook` into a standalone `mapGrafanaAlertToEvent(accountId string, alert map[string]any) (core.EventIncomingWebhook, error)` in `api-server/services/integrations/grafana_webhook.go`.

## Why

Per #257, an upcoming reconcile job that pulls alerts from Grafana's API needs the **exact same** alert→event mapping so both paths produce identical, `fingerprint`-keyed events (otherwise the two paths create duplicates instead of reconciling). This pulls the mapping into a single shared function both can reuse.

## Changes

- New `mapGrafanaAlertToEvent` holds the full per-alert mapping (fingerprint, firing/resolved status, start/end times, labels, severity, title, subject, tags, URLs).
- `ProcessEventWebook` now calls it inside its loop; the `CreateEventRule` fire-and-forget side effect stays in the loop.
- Behavior is unchanged.
- Adds `grafana_webhook_test.go` with unit tests covering a firing alert, title fallback to alertname, and dashboardURL fallback.

## Validation

- `gofmt -l` clean
- `go build ./integrations/` passes
- `go vet ./integrations/` clean
- `go test ./integrations/ -run TestMapGrafanaAlertToEvent` passes

Fixes #257

### DIFF
--- a/api-server/services/integrations/grafana_webhook.go
+++ b/api-server/services/integrations/grafana_webhook.go
@@ -130,7 +130,7 @@ func (m GrafanaWebhook) ProcessEventWebook(sc *security.RequestContext, settings
 			continue
 		}
 
-		results = append(results, parsedPayload)
+		results = append(results, *parsedPayload)
 	}
 
 	if len(results) == 0 {
@@ -146,7 +146,7 @@ func (m GrafanaWebhook) ProcessEventWebook(sc *security.RequestContext, settings
 // severity, title) so that both the inbound webhook path and any reconcile
 // job that pulls alerts from Grafana's API produce identical, fingerprint-keyed
 // events.
-func mapGrafanaAlertToEvent(accountId string, alert map[string]any) (core.EventIncomingWebhook, error) {
+func mapGrafanaAlertToEvent(accountId string, alert map[string]any) (*core.EventIncomingWebhook, error) {
 	labels := mapStringAnyToStringString(alert["labels"])
 	annotations := mapStringAnyToStringString(alert["annotations"])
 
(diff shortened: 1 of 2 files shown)


mayankpande88 (approved):
Nice and focused. The single mapGrafanaAlertToEvent mapper is exactly what the issue was after, the fingerprint hard-error is a sensible tightening, and the tests cover the firing/resolved/fallback/missing-fingerprint cases. Confirmed the existing mapGrafanaStatus/Severity helpers stay in place so the Prometheus path is unaffected. Thanks!

PR #270, opened 2026-06-20
refactor(integrations): extract reusable Grafana alert-to-event mapping

### LINKED ISSUE #257: refactor(integrations): extract reusable Grafana alert-to-event mapping

anupdubbewar:
Fix #257 

Move duplicated mapGrafanaStatus, mapGrafanaSeverity, mapStringAnyToStringString, and extractGrafanaSubject into shared alert_mapping.go with exported names. Both grafana_webhook.go and prometheus_alertmanager_wehbook.go now call the shared functions.

### DIFF
--- a/api-server/services/integrations/alert_mapping.go
+++ b/api-server/services/integrations/alert_mapping.go
@@ -1,0 +1,68 @@
+package integrations
+
+import (
+	"nudgebee/services/event"
+	"strings"
+)
+
+// MapAlertStatus maps Grafana/Prometheus lowercase status to EventStatus constants.
+func MapAlertStatus(status string) event.EventStatus {
+	switch strings.ToLower(strings.TrimSpace(status)) {
+	case "firing":
+		return event.EventStatusFiring
+	case "resolved":
+		return event.EventStatusResolved
+	case "closed":
+		return event.EventStatusClosed
+	default:
+		return event.EventStatusFiring
+	}
+}
+
+// MapAlertSeverity maps Grafana/Prometheus lowercase severity to EventPriortiy constants.
+func MapAlertSeverity(severity string) event.EventPriortiy {
+	switch strings.ToLower(strings.TrimSpace(severity)) {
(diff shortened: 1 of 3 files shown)


mayankpande88:
Heads up: #266 just merged, which extracts the per-alert Grafana mapping into mapGrafanaAlertToEvent with tests and covers the core of #257. This PR also carries an unrelated change to prometheus_alertmanager_wehbook.go that removes some explanatory comments. The shared-helper idea (sharing MapAlertStatus/Severity across the Grafana and Prometheus webhooks) is reasonable as a follow-up, but as-is this now overlaps with what merged. I would suggest closing this, or if you want to pursue the shared-helpers cleanup, rebase on main and drop the comment deletions. Your call.

PR #275, opened 2026-06-21
feat(notifications): Discord integration 

### LINKED ISSUE #255: feat(notifications): add Discord as an outbound notification channel
## Summary

Add **Discord** as an outbound notification channel, alongside the existing Slack, MS Teams, and Google Chat channels. NudgeBee findings/notifications should be deliverable to a Discord channel, configurable like the other platforms. (Inbound/interactive — slash commands, replies, "Ask NudgeBee in Discord" — is intentionally out of scope here.)

## Backend (`notifications-server`)

Follow the existing per-platform pattern:

- **Client**: add `notifications_server/clients/discord_client.py` (post a message to a channel). Recommended auth: a **bot token** so channels can be listed/selected like the other platforms; a per-channel **incoming webhook URL** is a simpler fallback but can't enumerate channels.
- **Message template**: add `notifications_server/message_templates/discord/finding.py` (mirror `slack/finding.py` / `ms_teams/finding.py`).
- **Formatting**: extend `utils/transformer.py` to render blocks to Discord (markdown + embeds) — reuse `to_standard_markdown` where possible.
- **Dispatch**: add a Discord sender path in `services/message.py` and register `discord` everywhere platforms are handled (`MessagingPlatform`, rule platform mappings, `list_channels` / `list_users` in `routers/common.py`).

## UI (`app`)

- Add `discord` to the platform list (the `slack | ms_teams | google_chat` set — see the `PlatformInput` Literal from #234) so it appears in channel/user listing and notification-rule channel selection.
- Add a Discord integration/config entry (bot token or webhook URL) + icon.

## Done when

- [ ] A Discord channel can be configured as a notification target
- [ ] A finding/notification is delivered to Discord, formatted (embed/markdown), at parity with the other channels
- [ ] `discord` is selectable in notification rules and channel listing
- [ ] Missing/invalid credentials produce a clear error; no inbound/interactive endpoint required

reckless-sherixx:
# Description

Added Discord as an outbound notification channel for NudgeBee findings, alongside the existing integrations. The integration provides the ability to post markdown-formatted alerts using Discord webhooks/bot tokens, and can list text channels for user configuration.

Fixes #255

## Type of change

- [x] New feature

# How Has This Been Tested?

- [x] Verified Discord message rendering using the standard embed formatting.
- [x] Tested channel listing validation to prevent unauthorized channel iteration.
- [x] Validated backend code using `poetry run black` and `poetry run flake8`.
- [x] Validated frontend code using `npm run lint2:fix`.

---
# Screenshots 

{5CF49898-5839-41C8-96A2-42E53A12DC29}

{FE86D13B-DA4D-4FA1-A54A-A10922F49209}

{543691D4-BEFA-43E1-AA53-467D2D338F8D}

image

{F5941E81-B987-45FF-9B16-029A339945FF}

{F2A830F4-991E-4F1E-AA3F-7DDFCDAAA317}

# Risks & Counterarguments

- **Risk**: Hardcoded conditional paths (`if platform == "discord"`) within `common.py` and `message.py`. 
  - **Why accepted**: It directly follows the pre-existing repository patterns for Slack, Teams, and Google Chat without introducing premature abstraction. 
  - **Revisit trigger**: If a generic webhook/plugin integration architecture is implemented later, all platforms including Discord should be abstracted.
- **Risk**: Returning `[]` for Discord user enumeration.
  - **Why accepted**: Discord bots require privileged Gateway intents to enumerate server members, which is out of scope for the notification bot's permissions. 
  - **Revisit trigger**: If direct message notifications to Discord members become necessary.


notifications-server/notifications_server/services/common.py
```diff
@@ -120,9 +122,24 @@ def get_slack_channels(self, messaging_platform):
 
         return {"data": channels}
 
+    def get_discord_channels(self, messaging_platform):
+        """List text channels the Discord bot has access to across all guilds."""
+        from notifications_server.clients.discord_client import DiscordClient
+
+        token = (
+            messaging_platform.credentials.get("bot_token")
```
saiprasadpotdar: `messaging_platform.credentials` isn't actually a field on the model.

If you look at `MessagingPlatform` in models.py, the columns are `token`, `channels`, `bot_id`, etc. - there's no `credentials`. And since `Base` is a plain `DeclarativeBase` (no `__getattr__`), `messaging_platform.credentials` just raises `AttributeError`.

It shows up in two places:

- `get_discord_channels`(~L130) 
- `_send_test_discord` (~L1903)

The bot token already lives in the `token` column (your own `acquire_discord_access_token` returns `ip.token`), so you can just drop the `credentials` in both spots:

```python
token = messaging_platform.token
```

notifications-server/notifications_server/services/message.py
```diff
@@ -899,6 +910,51 @@ async def _send_finding_to_gchat(self, session, tenant_id, ip, finding, _):
...
+        if not token:
+            return None, None
+
+        from notifications_server.clients.discord_client import DiscordClient
+        from notifications_server.message_templates.discord.finding import get_discord_finding_message
+
+        discord_payload = get_discord_finding_message(finding)
+
+        # ip.to_channel is set by get_channels_from_request_or_defaults
+        channel_id = ip.to_channel
```
saiprasadpotdar: I think findings will actually never make it to Discord here - `ip.to_channel` isn't a bare channel-id string at this point.

`get_channels_from_request_or_defaults` sets it to `rule["channels"]`, and the frontend persists that as `{ name, id }` (see NotificationRuleModal.tsx: `channels: { name: channel['label'], id: channel['value'] }`). `extract_channels` and `normalize_channel` both treat it as str(JSON)/list/dict too. So `chat_post` ends up building a URL like `https://discord.com/api/v10/channels/{'name': '…', 'id': '…'}/messages` - 404 every time.

The fallback right below (`ip.channels[0].get("id")`) doesn't save it either, since `send_finding_notification` already skips anything where `to_channel` is falsy (`if not ip.to_channel: continue`) - so `channel_id` is always a truthy dict and the fallback never runs.

This is similar to how `GoogleChatClient.post_to_google_chat` runs the channel through `_normalize_space()` internally. Easiest fix is to mirror gchat and normalize before the call:

```python
channel = normalize_channel(ip.to_channel)
channel_id = channel.get("id") if isinstance(channel, dict) else channel
```

saiprasadpotdar:
hey @reckless-sherixx 

Thanks for tackling this, nice work! The frontend wiring and the platform structure look great.

I've left a few inline comments. These ones are blockers rather than nits: as it stands I don't think findings actually reach Discord yet (the channel-id and `credentials` issues), so it'd be worth re-testing.

One bigger thing that doesn't map to a single line - I couldn't find an install/config flow for Discord. 

`account-form.jsx` has no `discord` case and there's no bot-token entry anywhere, so there's no way to actually connect Discord yet.

Marking as request-changes for now, but this is a solid start! 🙌

reckless-sherixx:
@saiprasadpotdar Thanks I'll fix all of them and raise a pr. I was having some bugs locally so I was unable to check everything. I just tried to clone the slack and google chat functionality.

mayankpande88 (approved):
Wiring looks complete: DiscordSender is registered in the dispatch map, channel listing and token resolution are in place, and chat_post runs via asyncio.to_thread so it does not block the event loop. Two small follow-ups for later, not blockers: the requests calls have no timeout, and channels_list does one API call per guild. Thanks for the thorough PR!

mayankpande88 (dismissed):
Retracting my earlier approval. I went back through the code and @saiprasadpotdar is right on both points, these are real and block merge:

1. `messaging_platform.credentials` does not exist on the model. `MessagingPlatform` only has `token`/`refresh_token`, so `messaging_platform.credentials.get("bot_token")` raises AttributeError in both `get_discord_channels` and `_send_test_discord`. Use `messaging_platform.token` (which is what your own `acquire_discord_access_token` already returns).

2. `ip.to_channel` is a `{name, id}` dict at this point, not a bare channel id, so `chat_post` builds `/channels/{'name':...,'id':...}/messages` and 404s every time. The other senders normalize first (gchat runs it through `_normalize_space`). Mirror that:
```python
channel = normalize_channel(ip.to_channel)
channel_id = channel.get("id") if isinstance(channel, dict) else channel
```
The `ip.channels[0]` fallback never runs because `send_finding_notification` already skips falsy `to_channel`.

The structure and async handling are good. Once these two are fixed and you have verified a finding actually lands in a Discord channel end to end, ping me and I will re-review.

notifications-server/notifications_server/routers/discord.py
```diff
@@ -0,0 +1,70 @@
...
+from notifications_server.clients.discord_client import DiscordClient
+from notifications_server.message_templates.base import render_success_page
+from notifications_server.configs.settings import settings
+from notifications_server.models.models import MessagingPlatform
+from notifications_server.repositories.oauth_repository import find_installation_by_tenant_and_platform
+from notifications_server.utils.datetime_utils import utc_now
+
+LOG = logging.getLogger(__name__)
+
+router = APIRouter(
```
saiprasadpotdar: This could go in `tools.py` rather than a new router? It's declaring the same `/api/integrations` prefix that `tools.py` already uses, and that's where the other integration installs live (`install/google`, `install/ms-teams`) - so `/install/discord` would slot right in next to them. Same prefix means the path doesn't change, so the frontend `install.ts` stays as-is too. A separate file mostly just spreads the install endpoints across two modules without buying us much.

If you do move it: `tools.py` already imports `Session`, `sync_engine`, and `find_installation_by_tenant_and_platform`, so it's mostly pasting the handler over, adding the `DiscordClient` / `MessagingPlatform` imports, and dropping the `include_router(discord.router)` line in `server.py`.

While you're in here - the Google flow persists via `service.save_google_chat_installation(...)`, but this builds the `MessagingPlatform()` and commits inline in the route handler. A matching `save_discord_installation(...)` would keep the persistence in the same place as the others.

How to make the change:
1. Server.py - drop discord from the import and remove the include
2. Move the handler into tools.py - paste DiscordInstallRequest + discord_install there.
3. Delete routers/discord.py

saiprasadpotdar (changes requested):
Really nice turnaround on this 🙌 All three of the earlier blockers are sorted!

What's left is minor: mostly a router-placement suggestion, added them inline. They're in routers/discord.py

saiprasadpotdar (approved):
Approving this 🎉

One non-blocking nit remains — the `sqlite3.IntegrityError` import — which I'll track as a separate follow-up issue rather than hold this up. 

Nice work! 🙌

reckless-sherixx:
Thanks @saiprasadpotdar assign me the remaining fix ill fix that too

PR #306, opened 2026-06-22
security(ui): validate relay proxy segment against endpoint allowlist

### LINKED ISSUE #287: 🔒 Harden relay API route: validate `[relay]` segment against an allowlist
## Summary

The relay proxy handler at `app/src/pages/api/proxy/relay/[relay].ts` interpolates the dynamic `relay` URL segment directly into the upstream fetch URL without validating it against an allowlist of known relay-server endpoints. This is a defense-in-depth hardening, not an unauthenticated vulnerability — auth and account-access gates already run before the unsafe concatenation.

## Context

Current code shape (around line 118):

```ts
const { relay } = req.query;              // dynamic Next.js segment
// ...
const response = await fetch(`${relayEndpoint}/${relay}`, {
  headers,
  body: JSON.stringify(req.body),
  method: 'post',
});
```

### Mitigations already in place

- **Auth gate** (around lines 44–85) — rejects with `401 not_authenticated` if there's no valid session / JWT. Unauthenticated callers cannot reach the concatenation.
- **Account-access gate** (around lines 93–98) — after auth, `hasAccountAccess(userId, tenantId, accountId, traceParent)` must return true; otherwise `403 forbidden`.
- Next.js dynamic-route segments (`[relay].ts` is a single-segment matcher) constrain `relay` to one URL path component, so a literal `/` in the request path routes elsewhere and never reaches this handler.

### What's still loose

- No allowlist of valid relay-server endpoints — any string matching the dynamic segment is forwarded.
- URL-encoded path separators (`%2F`, `%2e%2e%2f`) may or may not survive Next.js normalization across versions; the handler doesn't defend.
- Query-string can be appended via the segment value (`relay=foo?bar=evil`) — bounded by what relay-server's own route table actually accepts.

## Proposed fix

Right after `const { relay } = req.query;` (around line 16), validate against a static allowlist enumerated from the actual relay-server routes:

```ts
const ALLOWED_RELAY_PATHS = new Set(['request', /* … fill in from relay-server route table … */]);
const relay = String(req.query.relay ?? '');
if (!ALLOWED_RELAY_PATHS.has(relay)) {
  return res.status(400).json({ error: 'Invalid relay path' });
}
```

## Acceptance

- Allowlist enumerated from `collector-server/k8s-collector/relay-server` route definitions so no real caller breaks.
- Requests with unknown `relay` values return 400.
- Existing valid callers (e.g. `XtermTerminal`, ingest pushers) continue to work.

shivam2003-dev:
## Problem

`app/src/pages/api/proxy/relay/[relay].ts` interpolates the dynamic Next.js `[relay]` URL segment **directly** into the upstream fetch URL:

```ts
const response = await fetch(`${relayEndpoint}/${relay}`, { ... });
```

The segment is never checked against the set of endpoints the relay-server actually exposes. Auth + account-access gates run first (so this is defense-in-depth, not an unauthenticated hole), but an authenticated caller can still steer the request path to an arbitrary upstream path on the relay host.

## Fix

Validate the segment against an allowlist **before** forwarding:

```ts
const ALLOWED_RELAY_ENDPOINTS = new Set(['request', 'grafana', 'ws']);
// ...
const relaySegment = Array.isArray(relay) ? relay[0] : relay;
if (!relaySegment || !ALLOWED_RELAY_ENDPOINTS.has(relaySegment)) {
  res.status(400).json({ error: 'invalid_relay_endpoint', description: 'Unknown relay endpoint' });
  return;
}
```

The forward then uses the validated `relaySegment`.

### Why these three values

These are the only single-segment endpoints the app legitimately hits through this proxy:

- `request` and `grafana` — produced by `hitRelayServer()` in `app/src/lib/HttpService.ts` (`pathVariable = '/request'` default, `'/grafana'` for `type=grafana`).
- `ws` — registered on the relay-server router (`collector-server/k8s-collector/relay-server/pkg/server/router.go`, `r.POST("/ws", ...)`).

## How to verify

1. `POST /api/proxy/relay/request` (authenticated, valid account) → unchanged, forwards to `${relayEndpoint}/request`.
2. `POST /api/proxy/relay/anything-else` → now returns `400 invalid_relay_endpoint` instead of being forwarded.
3. No behavior change for the existing `request` / `grafana` callers.

Scope: single file, no API/contract change for legitimate callers.

Fixes #287

### DIFF
--- a/app/src/pages/api/proxy/relay/[relay].ts
+++ b/app/src/pages/api/proxy/relay/[relay].ts
@@ -11,6 +11,11 @@ import { context, propagation, trace, SpanStatusCode } from '@opentelemetry/api'
 const relayEndpoint = process.env.RELAY_SERVER_ENDPOINT ?? 'http://localhost:52832';
 const secretKey = process.env.RELAY_SERVER_SECRET_KEY ?? '';
 
+// Allowlist of known relay-server endpoints reachable through this single-segment
+// proxy. The dynamic `[relay]` segment is interpolated into the upstream fetch URL,
+// so it must be validated against this set to prevent forwarding to arbitrary paths.
+const ALLOWED_RELAY_ENDPOINTS = new Set(['request', 'grafana', 'ws']);
+
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   const tracer = trace.getTracer('relay-api');
   const { relay } = req.query;
@@ -97,7 +102,15 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         return;
       }
 
-      // --- Step 3: Relay request ---
+      // --- Step 3: Validate relay endpoint against allowlist ---
+      const relaySegment = Array.isArray(relay) ? relay[0] : relay;
+      if (!relaySegment || !ALLOWED_RELAY_ENDPOINTS.has(relaySegment)) {
+        span.setStatus({ code: SpanStatusCode.ERROR, message: 'Invalid relay endpoint' });
+        res.status(400).json({ error: 'invalid_relay_endpoint', description: 'Unknown relay endpoint' });
+        return;
+      }
+
+      // --- Step 4: Relay request ---
       const relaySpan = tracer.startSpan('relayFetch', undefined, trace.setSpan(context.active(), span));
       const headers: Record<string, string> = {
         'Content-Type': 'application/json',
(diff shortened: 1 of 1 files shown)


shivam2003-dev:
Thanks @gemini-code-assist — addressed both points in the latest commit:

1. **Removed `'ws'`** from the allowlist; the proxy only forwards HTTP POST, so the WebSocket route isn't reachable here. Allowlist is now `['request', 'grafana']` (the only segments `hitRelayServer()` produces).
2. **Moved validation to the top of the handler** (right after reading `req.query.relay`, before the span is created). This fails fast and keeps the untrusted segment out of the span name (`relay-handler-${relaySegment}` now uses the validated value), and removes the redundant mid-handler check.

mayankpande88 (approved):
Good defense-in-depth. Verified the allowlist {request, grafana} is exactly complete: hitRelayServer only ever uses /request, or /grafana when type=='grafana', so nothing legitimate is blocked. Validation sits at the very top before the span and fetch, which also keeps the untrusted segment out of the span name. The Gemini notes (no ws in the set, validate-at-top) are already satisfied by the current code. Thanks!

PR #308, opened 2026-06-22
Rename RAG reranking minimum docs setting

### LINKED ISSUE #173: bug(rag-server): reranking doc "cap" uses `max()` — `Config.reranking_max_docs` acts as a floor, not a ceiling
## Summary

`llm/rag-server/rag/core/llm/rag.py:137` computes how many documents to send to the LLM for reranking, with a comment saying "Cap docs sent to LLM to reduce input tokens." But it uses `max()`, which makes the config value a **lower bound** — the slice can only ever grow, never shrink.

The result: a large `no_of_results` sends `2 × no_of_results` docs to the LLM regardless of the configured maximum, contradicting the stated intent.

## The code

```python
# Sort by relevance score
similar_docs_flat.sort(key=lambda x: x[1], reverse=True)

if use_reranking:
    # Cap docs sent to LLM to reduce input tokens and latency.
    # Only send the top candidates by similarity for reranking.
    max_docs_for_reranking = max(no_of_results * 2, Config.reranking_max_docs)
    docs_for_reranking = similar_docs_flat[:max_docs_for_reranking]
    if len(similar_docs_flat) > max_docs_for_reranking:
        logger.info(f"Capping docs for LLM reranking: {len(similar_docs_flat)} -> {max_docs_for_reranking}")
```

Default config (`llm/rag-server/utils/config.py:62`): `Config.reranking_max_docs = int(os.environ.get("RAG_RERANKING_MAX_DOCS", 10))`.

So if `no_of_results = 50`, `max_docs_for_reranking` becomes `max(100, 10) = 100` — well above the documented cap of 10. The "Capping" log line will print but the value isn't actually a cap.

## Two possible intents — needs maintainer confirmation

This is a `status:needs-info` issue because the right fix depends on which of these the original author meant:

**(a) The comment is right and the code is wrong** — `Config.reranking_max_docs` is the hard ceiling. Fix:

```python
max_docs_for_reranking = min(no_of_results * 2, Config.reranking_max_docs)
```

This caps at the configured maximum (default 10) regardless of `no_of_results`.

**(b) The code is right and the comment is wrong** — `reranking_max_docs` is deliberately a floor, so reranking always sees at least N candidates even when the query asks for fewer results. Fix:

```python
# Send at least `reranking_max_docs` candidates to the reranker (floor), or
# 2× the requested results, whichever is larger.
max_docs_for_reranking = max(no_of_results * 2, Config.reranking_max_docs)
```

And rename the config var to something like `reranking_min_candidates` since it functions as a floor.

## What's needed

A short call from someone with context on the reranking flow:
- What is `Config.reranking_max_docs` intended to control?
- Should a config value named `_max_docs` ever be exceeded?

Once confirmed, the fix is a one-line change (and possibly a config rename + doc update).

## Why this matters

Token cost / latency, not correctness. The reranking LLM call still returns useful results — it just processes more documents than the operator configured, which is exactly what the cap was supposed to prevent.

## Severity

Medium. Affects every reranking call where `no_of_results * 2 > reranking_max_docs`.

shivam2003-dev:
Fixes #173.

## What changed
- Kept the reranking candidate calculation as `max(no_of_results * 2, Config.reranking_min_docs)`, preserving the intended 2x oversampling for large requests and the minimum pool for small requests.
- Renamed `Config.reranking_max_docs` to `Config.reranking_min_docs` so the code matches the confirmed floor semantics.
- Added `RAG_RERANKING_MIN_DOCS` while keeping `RAG_RERANKING_MAX_DOCS` as a backward-compatible fallback for existing deployments.
- Reworded the reranking comments and log message so they no longer describe the setting as a cap or ceiling.

## Root cause
The runtime behavior was intentional, but the setting name, environment variable, comment, and log text described a maximum. That made a floor look like a broken cap.

## Maintainer verification
1. `cd llm/rag-server`
2. `python3 -m black --check --target-version py313 rag/core/llm/rag.py utils/config.py`
3. `python3 -m compileall rag/core/llm/rag.py utils/config.py`
4. In an environment with dev dependencies installed, run `python3 -m flake8 rag/core/llm/rag.py utils/config.py`.

## Local checks
- PASS `python3 -m black --check --target-version py313 llm/rag-server/rag/core/llm/rag.py llm/rag-server/utils/config.py`
- PASS `python3 -m compileall llm/rag-server/rag/core/llm/rag.py llm/rag-server/utils/config.py`
- PASS `git diff --check`

Note: direct `python3 -m flake8 ...` could not run locally because `flake8` is not installed in this Python environment.

### DIFF
--- a/llm/rag-server/rag/core/llm/rag.py
+++ b/llm/rag-server/rag/core/llm/rag.py
@@ -132,12 +132,16 @@ def get_matching_documents(
         similar_docs_flat.sort(key=lambda x: x[1], reverse=True)
 
         if use_reranking:
-            # Cap docs sent to LLM to reduce input tokens and latency.
-            # Only send the top candidates by similarity for reranking.
-            max_docs_for_reranking = max(no_of_results * 2, Config.reranking_max_docs)
-            docs_for_reranking = similar_docs_flat[:max_docs_for_reranking]
-            if len(similar_docs_flat) > max_docs_for_reranking:
-                logger.info(f"Capping docs for LLM reranking: {len(similar_docs_flat)} -> {max_docs_for_reranking}")
+            # Send 2x the requested results so the LLM has room to reorder,
+            # while keeping a minimum candidate pool for small requests.
+            docs_for_reranking_count = max(no_of_results * 2, Config.reranking_min_docs)
+            docs_for_reranking = similar_docs_flat[:docs_for_reranking_count]
+            if len(similar_docs_flat) > docs_for_reranking_count:
+                logger.info(
+                    "Limiting docs for LLM reranking pool: %s -> %s",
+                    len(similar_docs_flat),
+                    docs_for_reranking_count,
+                )
 
             llm = get_llm(account_id)
             docs_reranked, token_usage = rerank_with_llm(query, module, docs_for_reranking, llm)
--- a/llm/rag-server/utils/config.py
+++ b/llm/rag-server/utils/config.py
@@ -59,7 +59,9 @@ class Config:
     rag_collection_cache_ttl = int(
         os.environ.get("RAG_COLLECTION_CACHE_TTL", 1800)
     )  # Qdrant collection list cache (default 30 min)
-    reranking_max_docs = int(os.environ.get("RAG_RERANKING_MAX_DOCS", 10))  # max docs sent to LLM for reranking
+    reranking_min_docs = int(
+        os.environ.get("RAG_RERANKING_MIN_DOCS", os.environ.get("RAG_RERANKING_MAX_DOCS", 10))
+    )  # minimum docs sent to LLM for reranking; old env name kept as fallback
     search_max_workers = int(
         os.environ.get("RAG_SEARCH_MAX_WORKERS", 8)
     )  # thread pool size for parallel collection search


mayankpande88:
The min() direction is right (it makes reranking_max_docs a real ceiling and fixes the token blowup in #173), but the Gemini high-priority note on this one is valid and worth handling before merge.

With min(), when no_of_results is larger than reranking_max_docs (e.g. k=50, cap=10), only the top 10 get sent to the reranker, and since the function returns docs_reranked[:no_of_results] at the end, the caller now gets at most 10 back instead of 50. The old max() avoided that by sending more docs to rerank. So this trades the token bug for a result-count regression on large-k callers. k defaults to 1 so most paths are fine, but k is caller-controllable.

The clean fix is to cap only what goes to the LLM, then append the remaining similarity-ordered docs back before the final slice, so you keep both the token cap and the requested count:

```python
max_docs_for_reranking = min(no_of_results * 2, Config.reranking_max_docs)
docs_for_reranking = similar_docs_flat[:max_docs_for_reranking]
if len(similar_docs_flat) > max_docs_for_reranking:
    logger.info(f"Capping docs for LLM reranking: {len(similar_docs_flat)} -> {max_docs_for_reranking}")
llm = get_llm(account_id)
docs_reranked, token_usage = rerank_with_llm(query, module, docs_for_reranking, llm)
docs_reranked = docs_reranked + similar_docs_flat[max_docs_for_reranking:]
```

That keeps #173 fixed without dropping results. Mind adding that?

shivam2003-dev:
Good catch — fixed in the latest commit. The `min()` cap correctly limits how many docs go to the LLM reranker, but as you noted, `docs_reranked[:no_of_results]` then truncated the final output to `reranking_max_docs` when a larger `no_of_results` was requested.

I now append the remaining similarity-sorted docs (the ones beyond the rerank cap) after the reranked top candidates before slicing:

```python
docs_reranked = docs_reranked + similar_docs_flat[max_docs_for_reranking:]
```

So only the top `reranking_max_docs` get LLM-reranked (the goal of #173), and the requested `no_of_results` is still fulfilled from the similarity-ordered tail.

shivam2003-dev:
Follow-up on the earlier review thread: the PR has been repurposed after maintainer clarification on #173.

The current branch no longer uses the `min()` cap approach. It keeps the intended `max(no_of_results * 2, Config.reranking_min_docs)` behavior, renames the setting to match the floor semantics, adds `RAG_RERANKING_MIN_DOCS` with the old env var as fallback, and updates the misleading comments/log text.

Current checks are green on the updated branch.

blue4209211 (changes requested):
make lint

black --check .
Warning: Python 3.11 cannot parse code formatted for Python 3.14. To fix this: run Black with Python 3.14, set --target-version to py311, or use --fast to skip the safety check. Black's safety check verifies equivalence by parsing the AST, which fails when the running Python is older than the target version.
All done! ✨ 🍰 ✨
59 files would be left unchanged.
flake8 .
mypy ./**/*.py --namespace-packages
controllers/migration_controller.py:311: error: List item 0 has incompatible type "dict[str, dict[str, Any]]"; expected "CreateAliasOperation | RenameAliasOperation | DeleteAliasOperation | AliasOperations"  [list-item]
Found 1 error in 1 file (checked 16 source files)
make: *** [lint] Error 1

PR #309, opened 2026-06-22
feat(ui): group Knowledge Graph account filter by cloud provider (#285)

### LINKED ISSUE #285: 🐛 [BUG] - Group Accounts by Cloud Provider in Knowledge Graph Account Filter
## Description

In the Knowledge Graph view, the **Account** filter currently lists every account in a flat dropdown. When the tenant has many accounts spread across providers (AWS, Azure, GCP, Kubernetes, etc.) this becomes hard to scan and select from, and it makes it easy to accidentally pick a selection large enough to hit graph-rendering limits.

## Proposed behavior

Group the Account filter entries by cloud provider (e.g. AWS, Azure, GCP, Kubernetes) — collapsible/header-separated — so the list is easier to navigate and the user can pick by provider first, then by account.

## Reproduction steps

1. Open the Knowledge Graph view.
2. Open the **Account** filter dropdown.
3. Observe a flat list of all accounts regardless of provider.

## Expected behavior

Accounts are visually grouped under their cloud provider, ideally with a per-group label and (optionally) a "Select all in group" affordance.

shivam2003-dev:
## Problem

On the Knowledge Graph view, the **Account** filter dropdown renders every cloud account in a single flat, ungrouped list. When a workspace has accounts across multiple cloud providers (AWS, Azure, GCP, K8s), the list becomes hard to scan and there is no visual separation by provider.

Fixes #285

## Fix

The shared `FilterDropdown` component (`@components1/ds/FilterDropdown`) already supports a `grouped` prop that renders option groups with headers, keying off an optional `group` field on each option. This PR wires that existing capability into the Knowledge Graph Account filter:

- `app/src/components1/KnowledgeGraph.jsx`
  - `accountOptions`: each option now carries `group: acc.cloud_provider`, so options are tagged with their provider (the cloud account objects already expose `cloud_provider`).
  - The Account `<FilterDropdown>` now passes the `grouped` prop, so the dropdown renders one section per cloud provider with a header.

No new dependencies, no API changes, and no changes to the shared `FilterDropdown` component — this only adds data and a flag the component already understands. Other filters (Node, Node Type, Level, etc.) are untouched.

## How to verify

1. Open the **Knowledge Graph** view in a workspace that has cloud accounts from more than one provider.
2. Open the **Account** filter dropdown.
3. Accounts are now grouped under provider headers (e.g. AWS, Azure, K8s) instead of a single flat list. Selecting/clearing accounts works exactly as before.

### DIFF
--- a/app/src/components1/KnowledgeGraph.jsx
+++ b/app/src/components1/KnowledgeGraph.jsx
@@ -1867,17 +1867,23 @@ const ServiceMapContent = () => {
     setPinnedNodeId(null);
   }, [clearPinHighlight]);
 
-  const accountOptions = useMemo(
-    () =>
+  const accountOptions = useMemo(() => {
+    const providerGroup = (provider) => {
+      if (!provider) return undefined;
+      const labels = { aws: 'AWS', gcp: 'GCP', azure: 'Azure', oci: 'OCI', k8s: 'Kubernetes' };
+      const key = provider.toLowerCase();
+      return labels[key] || provider;
+    };
+    return (
       accounts
         ?.filter((ac) => kgFilterOptions?.accountIds?.includes(ac.id))
         ?.map((acc) => ({
           label: acc.label || acc.account_name,
           value: acc.id || acc.value,
-          group: acc.cloud_provider,
-        })) || accounts,
-    [accounts, kgFilterOptions?.accountIds]
-  );
+          group: providerGroup(acc.cloud_provider),
+        })) || []
+    );
+  }, [accounts, kgFilterOptions?.accountIds]);
 
   return (
     <>
--- a/app/src/components1/KnowledgeGraph.jsx
+++ b/app/src/components1/KnowledgeGraph.jsx
@@ -1874,6 +1874,7 @@ const ServiceMapContent = () => {
         ?.map((acc) => ({
           label: acc.label || acc.account_name,
           value: acc.id || acc.value,
+          group: acc.cloud_provider,
         })) || accounts,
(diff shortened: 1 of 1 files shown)


mayankpande88:
The wiring is correct, accounts carry cloud_provider and FilterDropdown groups on opt.group, so this works. One small polish before merge: the group headers will render as 'Aws' / 'Gcp' / 'K8s' since FilterDropdown runs the raw value through snakeToTitleCase. Could you format them to the usual 'AWS' / 'GCP' / 'Kubernetes' so the headers match the rest of the app. While you are there, the || accounts fallback is unreachable (accounts starts as []) and would pass wrong-shaped options if it ever hit, so || [] is safer. Both are quick. Thanks!

shivam2003-dev:
Thanks for the review! Addressed both points in the latest commit:

1. **Group casing** — added an inline `providerGroup()` mapping that normalizes `cloud_provider` to display labels (`AWS`, `GCP`, `Azure`, `OCI`, `Kubernetes`). I kept it local rather than calling `snakeToTitleCase`, since that helper is not exported from `FilterDropdown.jsx` and has no acronym map (it would render `Aws`/`Gcp`/`K8s`). The map is case-insensitive (`provider.toLowerCase()`) and falls back to the raw value for any unknown provider.
2. **Fragile fallback** — changed `|| accounts` to `|| []` so we never hand raw account objects (missing `label`/`value`) to `FilterDropdown`.

RamanKharchee (approved):

saiprasadpotdar (approved):
LGTM ✅
Approving. 🚀

PR #332, opened 2026-06-23
test(ticket-server): add unit tests for common/secrets.go

bharathnr1:
# Description

`ticket-server/common/secrets.go` had no test coverage. This adds unit tests for:

- `Encrypt`/`Decrypt` AES-GCM round-trip (empty, plain, and symbol-heavy strings)
- `Decrypt` stripping the `enc:` prefix
- random-nonce uniqueness (repeated `Encrypt` yields distinct ciphertexts)
- the too-short-ciphertext error path
- `Encode` SHA-1 digest (known vector + determinism)

The encryption key is set via the package `Config` and restored with `t.Cleanup`. No production code changed.

Fixes #331

## Type of change

- [x] Chore (non-breaking change which does not modify src or test files)

# How Has This Been Tested?

- [x] `go test ./common/` — all 5 tests pass
- [x] `gofmt -l` clean

PR #334, opened 2026-06-23
fix(ticket-server): rewrite Encode determinism check to avoid SA4000

blue4209211:
## What

The new `TestEncode` test added in #332 has a `staticcheck SA4000` violation at `ticket-server/common/secrets_test.go:22`:

\`\`\`go
if Encode(\"x\") != Encode(\"x\") {  // identical expressions on both sides → SA4000
    t.Error(\"Encode is not deterministic for the same input\")
}
\`\`\`

\`golangci-lint\` fails on this on **every PR opened against \`main\`** because GitHub builds the PR's merge into \`main\` and the new file enters lint scope. Currently blocking the EE→OSS sync PR #305 (and any other PR landing after #332).

## Fix

Assign the two `Encode("x")` calls to locals first — staticcheck is happy and the test still asserts what it intended (same input → same digest):

\`\`\`go
first := Encode(\"x\")
second := Encode(\"x\")
if first != second {
    t.Error(\"Encode is not deterministic for the same input\")
}
\`\`\`

## Verification

\`\`\`bash
cd ticket-server
go test ./common/ -run TestEncode      # PASS
golangci-lint run ./common/...         # 0 issues
\`\`\`

### DIFF
--- a/ticket-server/common/secrets_test.go
+++ b/ticket-server/common/secrets_test.go
@@ -18,8 +18,12 @@ func TestEncode(t *testing.T) {
 	if got := Encode("hello"); got != want {
 		t.Errorf("Encode(\"hello\") = %q, want %q", got, want)
 	}
-	// Encode is deterministic.
-	if Encode("x") != Encode("x") {
+	// Encode is deterministic — calling it twice on the same input yields the
+	// same digest. Assigning to locals first avoids staticcheck SA4000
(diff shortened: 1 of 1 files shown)


mayankpande88 (approved):

mayankpande88 (approved):
Confirmed this clears the SA4000 failure and ticket-server app-build is green. Thanks for the quick fix.

PR #362, opened 2026-06-23
ci: fix benchmark/relay workflow path filters + the benchmark dep conflict it surfaces

bharathnr1:
# Description

Fixes path-filter correctness bugs found during a CI review:

- **`benchmark-server.yaml`** — the trigger path `llm/benchmark/server/**` matches **no code** (the module is at `llm/benchmark/app.py` + `llm/benchmark/benchmark_server/**`), so benchmark CI never ran on benchmark changes. Repointed to `llm/benchmark/**`.
- **Dead self-references** — editing a workflow should re-trigger it, but `benchmark-server.yaml` referenced a non-existent `benchmark-server-dev-civo.yaml` and `relay-server.yaml` referenced a non-existent `relay-server-test.yaml`. Pointed both at their real filenames.

Fixes #361

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Build / CI (non-breaking change which does not modify src or test files)

# How Has This Been Tested?

- [x] Both workflow files validated to parse (`yaml.safe_load`)
- [x] Confirmed the new `llm/benchmark/**` path covers the actual module layout (`app.py`, `benchmark_server/`)

## Review Notes → Risks & Counterarguments

- Fixing the benchmark path means its CI will now **actually run** on benchmark PRs (it was silently skipped before). If that job has latent failures that were never surfaced, they'll appear now — which is the point, but worth flagging.

Also fixes #363 (benchmark ragas/datasets dependency conflict — folded in from the now-closed #364).

PR #364, opened 2026-06-23
fix(benchmark): align ragas pin with the lock (0.4.3 -> 0.3.0rc2)

### LINKED ISSUE #363: fix(benchmark): unsatisfiable ragas/datasets pins break uv resolution
`llm/benchmark/pyproject.toml` pins `ragas==0.4.3` and `datasets==3.6.0`. ragas 0.4.3 requires `datasets>=4.0.0`, so `uv pip install -e .` fails:

```
Because ragas==0.4.3 depends on datasets>=4.0.0 and benchmark==0.1.0
depends on datasets==3.6.0, we can conclude that benchmark==0.1.0 and
ragas==0.4.3 are incompatible.
```

The conflict was latent because the benchmark CI never ran (its path filter `llm/benchmark/server/**` matched no code; fixed in #362). The committed `uv.lock` already resolves `ragas==0.3.0rc2` + `datasets==3.6.0`, and the code imports a private 0.3-era ragas API, so the intended/validated version is `0.3.0rc2`. Realign the pyproject pin to the lock.

bharathnr1:
# Description

`llm/benchmark/pyproject.toml` pinned `ragas==0.4.3` alongside `datasets==3.6.0`, but ragas 0.4.3 requires `datasets>=4.0.0` — an unsatisfiable combination that breaks `uv pip install -e .`:

```
Because ragas==0.4.3 depends on datasets>=4.0.0 and benchmark==0.1.0 depends on
datasets==3.6.0 ... benchmark==0.1.0 cannot be used.
```

The conflict was **latent** — the benchmark CI never ran (its path filter matched no code until #362 fixed it), so the bad pin was never exercised.

**Fix:** revert `ragas` to `0.3.0rc2`, the version the committed `uv.lock` already resolves and that the code was written against (it imports the private `ragas.metrics._domain_specific_rubrics` API, a 0.3-era surface). No `uv.lock` change needed — it already matches.

Verified with `uv pip compile pyproject.toml --python 3.13`: resolves cleanly (236 packages, exit 0).

Fixes #363

## Type of change

# How Has This Been Tested?

- [x] `uv pip compile pyproject.toml --python 3.13` resolves cleanly (236 packages) — previously errored on the ragas/datasets conflict
- [x] Confirmed the committed `uv.lock` already pins `ragas==0.3.0rc2` + `datasets==3.6.0` (no lock change required)

- **Why not keep ragas 0.4.3 and bump datasets to 4.x?** That's a real upgrade with breaking-change surface (datasets 4.0) and forces a full lock recompile that drifts dozens of transitive deps (aiohttp, anthropic, boto3, google-*). It should be a deliberate, separately-validated upgrade — not bundled into a CI-unblock. This PR restores the last-validated, consistent state instead.
- Relationship to #362: #362 (path-filter fix) is what makes the benchmark CI run and surfaced this. This PR should merge **before** #362 (or #362 rebased after) so #362's benchmark job goes green.

### DIFF
--- a/llm/benchmark/pyproject.toml
+++ b/llm/benchmark/pyproject.toml
@@ -143,7 +143,7 @@ dependencies = [
     "python-multipart>=0.0.18",
     "pytz==2026.1.post1",
     "pyyaml==6.0.3",
-    "ragas==0.4.3",
+    "ragas==0.3.0rc2",
     "regex==2024.11.6",
     "requests>=2.32.5",
     "requests-toolbelt==1.0.0",


bharathnr1:
@mayankpande88 PTAL.

bharathnr1:
Superseded by #362 — folded this one-line ragas pin fix directly into the CI path-filter PR so #362's benchmark job passes standalone (no merge-order dependency). The branch fix/benchmark-ragas-pin can be deleted.

PR #366, opened 2026-06-23
fix(api-server): close result set and surface iteration error in resolution lists

### LINKED ISSUE #365: 🐛 [BUG] - Resolution list helpers leak the result set and swallow iteration errors
## Pre-flight

- [x] I've reproduced on the latest `main`.

## Affected component(s)

api-server, services-server

## Version / commit

`main@5027933`

## Description

`ListEventResolutions` (`api-server/services/event/service.go:194`) and `ListRecommendationResolutions` (`api-server/services/recommendation/service.go:44`) each run `databaseManager.Db.Queryx(...)`, iterate the result with `for r.Next()`, but **never `defer r.Close()`** and **never check `r.Err()`** after the loop.

## Expected behavior

- The `*sqlx.Rows` result set is closed on every return path.
- A mid-iteration driver error is returned to the caller, not swallowed.

## Actual behavior

1. **Result-set leak.** On the `StructScan` error branch the function returns with the rows still open. `database/sql` auto-closes a `Rows` only when `Next()` reaches `io.EOF` or hits a driver error; an early return *mid-iteration* leaves the connection unreleased for the process lifetime.
2. **Swallowed iteration error / silent truncation.** If `Next()` stops because of a driver error (rather than normal end-of-rows), the loop ends and — with `r.Err()` unchecked — the function returns a **partial** slice and a `nil` error. For `ListRecommendationResolutions` this query backs the duplicate-resolution safety check (`status = 'InProgress' AND created_at > NOW() - INTERVAL '2 hours'`), so a truncated result can miss an in-progress row and allow a **duplicate resolution to be created**.

## Reproduction steps

1. Drive the helper with a result set whose row scan fails (e.g. an unexpected column) → the connection is not released.
2. Drive it with a result set that surfaces an error mid-iteration → the caller receives a truncated list with a `nil` error.

Both paths are covered deterministically by added `go-sqlmock` unit tests that fail on the current code and pass after the fix.

## Frequency

Always (deterministic at the code level).

PoorneshThatturi:
# Description

`ListEventResolutions` (`event/service.go`) and `ListRecommendationResolutions` (`recommendation/service.go`) iterated a `*sqlx.Rows` without closing it and without checking `rows.Err()`. Two defects followed:

1. **Result-set leak** — on the `StructScan` error branch the function returns with the rows still open. `database/sql` auto-closes a `Rows` only on `io.EOF`/driver error from `Next()`, so an early return mid-iteration leaks the connection for the process lifetime.
2. **Swallowed iteration error** — a driver error that ends iteration left `rows.Err()` unchecked, so the caller got a silently **truncated** slice with a `nil` error. `ListRecommendationResolutions` backs the duplicate-resolution safety check, so a truncated result can miss an in-progress row and allow a **duplicate resolution to be created**.

Fix (matches the existing idiom in `recommendation/exporter_*.go` and `tenant/service.go`):

- `defer func() { _ = r.Close() }()` immediately after the query.
- `if err := r.Err(); err != nil { return ..., fmt.Errorf("error iterating ... rows: %w", err) }` after the loop.

Fixes #365

## Type of change

# How Has This Been Tested?

- [x] New deterministic `go-sqlmock` unit tests for both helpers (run with `-race`), covering the close-on-scan-error path and the mid-iteration-error path. Each test **fails on the pre-fix code** and passes after the fix.
- [x] `go test -race ./event/ ./recommendation/` — added tests plus the existing package suites pass.
- [x] `gofmt` and `go vet` clean; `golangci-lint` at the CI-pinned `v2.11.3` reports **0 issues** on the module.

### DIFF
--- a/api-server/services/event/list_event_resolutions_test.go
+++ b/api-server/services/event/list_event_resolutions_test.go
@@ -1,0 +1,88 @@
+package event
+
+import (
+	"database/sql/driver"
+	"errors"
+	"nudgebee/services/internal/database"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/jmoiron/sqlx"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// eventResolutionColumns mirrors the column list selected by
+// ListEventResolutions.
+var eventResolutionColumns = []string{
+	"id", "created_at", "updated_at", "event_id", "type", "data", "status",
+	"type_reference_id", "resolver_type", "resolver_id", "status_message",
+	"pr_iteration_count", "pr_lifecycle_state", "last_pr_check_at",
+}
+
+// TestListEventResolutions is a regression test for two defects in
+// ListEventResolutions:
+//
+//  1. the result set was never closed on the scan-error early-return path, so a
+//     StructScan failure leaked the underlying connection for the process
+//     lifetime; and
+//  2. the loop never checked rows.Err(), so a mid-iteration driver error was
+//     swallowed and the caller received a silently truncated slice with a nil
+//     error.
+//
+// Both sub-tests drive the helper with a sqlmock-backed Metastore, so they run
+// with no live database.
+func TestListEventResolutions(t *testing.T) {
+	manager := &database.DatabaseManager{}
+	database.RegisterDatabaseManagerHook(database.Metastore, func() (*database.DatabaseManager, error) {
+		return manager, nil
+	})
+
+	// newMock swaps a fresh sqlmock-backed connection onto the manager the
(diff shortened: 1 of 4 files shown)


mayankpande88 (approved):
Verified locally: both bugs (result-set leak on the StructScan early-return + swallowed `rows.Err()`) are real in base, the fix matches the existing `tenant/service.go` / exporter idiom, and `fmt` is already imported in both files. The duplicate-resolution impact is genuine — `ListRecommendationResolutions`/`ListEventResolutions` gate `hasInProgressRecommendation`, so a truncated slice could allow a duplicate resolution. Ran `go test -race ./event/ ./recommendation/` → green; the new sqlmock regression tests assert close-on-scan-error and surfaced mid-iteration error correctly. LGTM.

PR #371, opened 2026-06-23
fix(llm): singleflight google ai cache creation

### LINKED ISSUE #302: [BUG] - LLM cache: redundant Google AI cache creations under concurrent load (singleflight needed)
## Description

Concurrent calls to `GoogleAICacheProvider.ApplyCache` for the **same `cacheKey`** race past the in-memory `CacheGet` check (around `llm/llm-server/agents/core/llm_cache.go:393`), all fall through to `createCache` (around `llm_cache.go:489`), and each creates a **distinct Google AI `cachedContents/...` resource**.

Only the last write wins in the local/Redis cache pointer, but every redundant Google AI resource keeps consuming storage cost (~\$4.50/M tokens/hr for Gemini Pro-class models) for its full TTL until expiry.

For an account-scoped cache key like `account:{accountId}:k8s_debug:gemini-pro`, all conversations within the account should share **one** Google AI resource. Currently they don't — under benchmark / parallel-conversation load we observe N redundant resources for the same key.

This is a **pre-existing bug** — the race window has existed since multi-level cache scoping was introduced. It became newly visible after the `llm_cache_lifecycle` tracking table (migration V727) landed: every redundant Google AI write now produces a row, making the waste observable in DB.

## Reproduction

1. Run a benchmark or any flow that spawns multiple parallel conversations against the same account hitting an account-scoped agent (e.g. `k8s_debug`).
2. Query `llm_cache_lifecycle`:

```sql
SELECT scope, agent_name, COUNT(*) AS rows, COUNT(DISTINCT conversation_id) AS distinct_convos
FROM llm_cache_lifecycle
WHERE created_at >= NOW() - INTERVAL '1 hour'
GROUP BY scope, agent_name;
```

3. Observe **N rows for the same `account:agent:model` key** — one per concurrent conversation that races into `createCache`. Expected: **1 row** regardless of concurrency.

Observed in a benchmark run: 12 lifecycle rows for `k8s_debug` across 6 distinct conversations (~2 redundant resources per conversation pair).

### Log shape

```
12:00:00.859  conversation A  k8s_debug  cached_tokens=23001  (cacheName A)
12:00:01.462  conversation B  k8s_debug  cached_tokens=23001  (cacheName B)  <- duplicate
12:04:05.569  conversation C  k8s_debug  cached_tokens=23001  (cacheName C)  <- duplicate
12:04:26.842  conversation D  k8s_debug  cached_tokens=23001  (cacheName D)  <- duplicate
…
```

## Proposed fix

Two-tier mitigation in `llm/llm-server/agents/core/llm_cache.go`:

1. **Tier 1 (in-process, mandatory)** — wrap the `createCache` call site (around line 489) in `golang.org/x/sync/singleflight.Group.Do(cacheKey, fn)`. Re-check `common.CacheGet(p.namespace, cacheKey)` at the top of `fn` to handle the small window where another goroutine just committed before us.

2. **Tier 2 (cross-replica, optional)** — add a Redis `SETNX`-based distributed lock keyed on `cacheKey` for prod multi-replica safety. Only relevant when llm-server runs &gt; 1 replica with shared Redis cache provider.

Tier 1 alone covers all observed reproductions (single-replica local + benchmark). Tier 2 is worth doing if multi-replica races are observed in prod.

## Environment

- **Service**: `llm-server` (Go)
- **Cache provider**: `bigcache` (in-memory, default) or Redis (multi-replica)
- **Affected provider**: GoogleAI only — Anthropic uses inline content-addressed `cache_control` and is unaffected.

## Cost impact

Each redundant cache resource costs `cached_tokens × storage_rate × ttl_hours`. For a Gemini Pro-class model at \$4.50/M/hr × 12h × 23,001 tokens ≈ **\$1.24 per orphaned cache**. With 11 redundant caches in the observed benchmark window, **~\$13.69 of storage waste in a single test run.** Scales linearly with parallel conversation count and benchmark frequency.

## Acceptance

- Under concurrent `ApplyCache` calls for the same `cacheKey`, **exactly one** `createCache` is issued and **exactly one** `llm_cache_lifecycle` row is written.
- A regression test in `llm/llm-server/agents/core/llm_cache_test.go` that spawns N goroutines on the same key and asserts a single creation.
- (Optional) opt-in Redis SETNX gate behind config for multi-replica deployments.

snvtac:
## Summary
- add an in-process singleflight gate around Google AI cached content creation for identical cache key + content hash misses
- re-check the shared cache inside the singleflight callback before creating provider-side cached content
- avoid charging cache creation tokens to singleflight waiters while still returning the cached content option
- add concurrency regression tests for same-content coalescing and different-content isolation

## Root cause
Concurrent `GoogleAICacheProvider.ApplyCache` calls could all pass the shared cache lookup before any goroutine stored the newly-created Google AI cached content pointer. Each caller then created a separate `cachedContents/...` resource for the same stable account/global cache content.

## Validation
- `go test ./agents/core -run 'TestGoogleAICacheProvider_(ApplyCacheSingleflightsConcurrentCreates|GetOrCreateCacheDoesNotSingleflightDifferentContent)' -count=1`
- `go test ./agents/core -run 'Test(Cache|GoogleAICacheProvider|GenerateCacheKey|AnthropicCacheProvider|GetLlmMaxTokenLength|ResolveLLMConfig_PerRequestCacheKeyedByTier)' -count=1`

Fixes #302

### DIFF
--- a/llm/llm-server/agents/core/llm_cache.go
+++ b/llm/llm-server/agents/core/llm_cache.go
@@ -18,6 +18,7 @@ import (
 	toolcore "nudgebee/llm/tools/core"
 
 	"github.com/tmc/langchaingo/llms"
+	"golang.org/x/sync/singleflight"
 )
 
 // CacheScope defines the stability level of the cache
@@ -160,7 +161,24 @@ const GoogleAICacheNamespace = "llm_googleai_cache"
 
 // GoogleAICacheProvider implements caching for Google AI (pre-created cached content)
 type GoogleAICacheProvider struct {
-	namespace string
+	namespace        string
+	cacheCreateGroup singleflight.Group
+
+	countTokensFunc       googleAICountTokensFunc
+	createCacheFunc       googleAICreateCacheFunc
+	verifyCacheExistsFunc googleAIVerifyCacheExistsFunc
+	deleteCacheFunc       googleAIDeleteCacheFunc
+}
+
+type googleAICountTokensFunc func(ctx context.Context, apiKey string, model string, messages []llms.MessageContent) (int32, error)
+type googleAICreateCacheFunc func(ctx context.Context, req *CacheRequest, cacheableMessages []llms.MessageContent, contentHash, cacheKey string, tokenCount int32) (*CacheInfo, error)
+type googleAIVerifyCacheExistsFunc func(ctx context.Context, cacheName, apiKey string) bool
+type googleAIDeleteCacheFunc func(ctx context.Context, cacheName, apiKey string) error
+
+type googleAICacheCreateResult struct {
+	cacheInfo *CacheInfo
+	created   bool
+	owner     *int
 }
 
 type CacheInfo struct {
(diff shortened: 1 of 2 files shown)


hsundar-git:
Thanks for tackling #302 🙏 — heads-up that there are **three open PRs for this same issue** (#371, #373, #387), all currently implementing **Tier 1 (`singleflight`) only**.

Two updates:

1. **Scope changed:** Tier 2 is now **required**, not optional — see #302. Prod runs **multiple replicas**, so `singleflight` (in-process) alone still creates one duplicate cache *per replica*. A complete fix needs the **cross-replica lock** (Redis `SETNX` on `cacheKey`) as well.
2. **How we'll pick one:** since these overlap, the **first PR that is fully _ready for review_** wins — meaning: complete scope (Tier 1 **and** Tier 2), **CLA signed**, CI green, and review comments addressed. Not "first opened" — _first actually ready_.

Please update your PR to cover Tier 2 (or coordinate among yourselves). Appreciate the work! 🚀

mayankpande88:
Closing as a duplicate of #387, which was just merged for #302. #387 implements the same singleflight dedup plus a cross-replica Redis lock. Thanks for the work.

PR #375, opened 2026-06-23
perf: optimize triage pipeline performance (Fixes #286)

### LINKED ISSUE #286: perf(api-server): N+1 queries in triage scoring + unbounded goroutines for rule-match updates
## Summary

Three related hot-path performance issues in the event triage pipeline in `api-server/services/triage/`. They are independent fixes but share the same per-event evaluation path.

### 1. N+1 queries in `ComputeScore()`

**File:** `api-server/services/triage/scoring.go` — `ComputeScore` (around line 66)

`ComputeScore` issues 3 separate DB queries per event:

- `getEnvironmentMultiplier` (~ line 79, defn ~ line 181) → `account_env`
- `getDuplicatePenalty` (~ line 91, defn ~ line 205) → `event_duplicates`
- `getCorrelationAdjustment` (~ line 102, defn ~ line 233) → correlation data

At 1000 events you'd issue ~3000 queries; at 10ms/query that's ~30s of pure DB round-trip.

### 2. Unbounded goroutines for rule-match counter updates

**File:** `api-server/services/triage/rules.go` — lines ~60, ~77, ~90

Each matched rule (one of Suppression / Scoring / Classification) fires:

```go
go updateRuleMatchCount(ctx, db, rule.ID)
```

with no semaphore, no worker pool, no batching on the hot per-event path. At 10K matching events you can spawn 10K goroutines competing for the bounded DB connection pool — pool exhaustion at scale.

A batched variant `updateRuleMatchCountBy(ctx, db, ruleID, count)` already exists (around line 648) but is only wired into the bulk-reclassify path in `bulk_operations.go`. The per-event path still uses the single-rule goroutine pattern.

### 3. No cache for `LoadMatchingRules`

**File:** `api-server/services/triage/rules.go` — `LoadMatchingRules` (around line 145), single caller `CheckTriageRules` (around line 19)

Each event triggers a fresh `SELECT … FROM event_triage_rules WHERE …`. Triage rules change rarely; this is a textbook TTL+invalidate cache candidate.

A separate `regexCache` (`sync.Map`, around line 696) already caches compiled regex patterns — that work is orthogonal to caching the rule list itself.

## Suggested fix order

1. **(1) + (2) together** — they share the per-event hot path. Batch the scoring lookups across an event batch and switch the rule-match counter to a worker-pool (or accumulator + single `UPDATE … WHERE id = ANY($1)`) at flush time.
2. **(3)** — TTL+invalidate cache on `event_triage_rules` CRUD. Lower risk because rule CRUD is rare.

## Acceptance

- No per-event goroutine spawn for rule-match counter updates under normal load.
- `LoadMatchingRules` does not hit DB on the per-event path more than once per cache TTL.
- A benchmark / load test demonstrating reduced query count at ≥1K events/batch.

reckless-sherixx:
# perf: optimize triage pipeline performance

Fixes #286

## Description
This PR addresses three critical hot-path performance bottlenecks in the event triage pipeline (`api-server/services/triage/`) to ensure high throughput processing at scale, while also reinforcing multi-tenant isolation and concurrent safety.

### 1. Eliminated N+1 Queries in `ComputeScore`
- Moved DB lookups upstream into `ProcessEvent` before scoring is triggered.
- Modified `detectAndRecordDuplicate` and `detectAndRecordCorrelations` to return `occurrenceNumber`, `correlationScore`, and `correlationType` directly instead of forcing `ComputeScore` to re-fetch them.
- Implemented a fast path 5-minute TTL cache for `getEnvironmentMultiplier` to avoid repeating queries against `cloud_accounts` per event (includes a fast-return for empty account IDs).
- **Behavioral Fix**: Recurring-fingerprint events now get a correlation adjustment correctly factored in alongside the duplicate penalty (previously omitted).

### 2. Fixed Unbounded Goroutines in `rules.go`
- Replaced the unbounded `go updateRuleMatchCount(ctx, db, rule.ID)` spawned per event match.
- Introduced a buffered package-level channel `matchCountChan` to enqueue match increments.
- Added a background daemon `runRuleMatchBatcher` that aggregates updates and flushes them to the DB efficiently every 5 seconds.
- **Resiliency**: Added a `recover()` block and a `ctx.Done()` flush loop to the background batcher daemon to ensure graceful shutdown and panic resilience.

### 3. Cached `LoadMatchingRules`
- Added an in-memory `triageRulesCache` bounded by a read/write mutex.
- Implemented a 60-second TTL cache for the tenant and account specific matching rules to completely offload read traffic on the `event_triage_rules` table during burst workloads.
- **Data Race Fix**: The cache now returns a deep copy of the slice (`rulesCopy`) to prevent concurrent mutation by callers.
- **Cache Invalidation**: The cache is properly invalidated (`ClearTriageRulesCache`) whenever rules are mutated via `CreateTriageRule`, `UpdateTriageRule`, `DeleteTriageRule`, `ToggleSystemRuleOverride`, or during auto-classification rule inserts in `ClassifyEvent`.

### 4. Enforced Strict Multi-Tenant Isolation
- Added `tenantID == ""` fail-fast guards in `LoadMatchingRules`, `detectAndRecordDuplicate`, and `detectAndRecordCorrelations` to prevent cross-tenant data leaks and to skip cross-event state operations on malformed events.

## Verification
- Confirmed unit tests pass with the refactored dependencies and decoupled function signatures.
- Tested locally without race conditions in the background rule matching aggregator.


mayankpande88 (commented):
Nice work covering all three parts of #286. I traced the refactor and the core is correct: the occurrence number threading matches what the old query returned, the correlation type is consistent across both directional rows, and the channel plus batcher removes the unbounded goroutine spawn.

One blocker: the new `triageRulesCache` is never invalidated. None of the rule mutation paths (`CreateTriageRule`, `UpdateTriageRule`, `DeleteTriageRule`, the two `classification.go` inserts) clear it, so a new or deleted suppression rule takes up to 60s to apply, and a deleted one can briefly hide alerts. System rules (tenant and account `NULL`) also live in every tenant's cached list, so a system rule change has to flush the whole map. Since rule edits are rare, the simplest fix is to clear the entire cache on any rule CRUD.

Two smaller things:
- Recurring-fingerprint events now get a correlation adjustment they didn't before. It's probably more correct, but worth a line in the PR description.
- The new tenant required-field guards skip dedup and correlation for events without a tenant. Intentional?

Heads-up: this overlaps #388 on `rules.go`. One option is to let #388 land first, then rebase and drop the batcher rewrite, keeping the N+1 fix and the cache with invalidation.

reckless-sherixx:
@mayankpande88 Yes, the new tenant required-field guards intentional defense-in-depth measure. Because event_duplicates and event_correlations operate across events, processing an event with an empty or missing tenant_id creates a risk of cross-tenant data bleed or polluting a global namespace. By failing fast, we ensure malformed events are strictly isolated and don't affect triage state. 

I have also included about Recurring-fingerprint events in the pr description and I think rebase is not needed and this pr fixed everything

mayankpande88 (approved):

PR #377, opened 2026-06-23
fix(notifications): catch SQLAlchemy IntegrityError on discord install (#376)

### LINKED ISSUE #376: [BUG] - Discord integration: duplicate-install error handling doesn't work (latent server-error risk)
## Description
The Discord notification integration has error handling that's supposed to catch the case where an installation can't be saved because it conflicts with an existing one, and turn it into a clear "installation already exists" message. That handler doesn't actually work — it watches for the wrong kind of database error, so it can never trigger. Today this has **no visible effect**: the normal "install Discord" flow works, and duplicate installs are already blocked earlier with a clear message. The risk is latent — if a database-level uniqueness rule is later added for integrations, a conflicting or racing Discord install would surface as a generic **500 server error** instead of the intended clear message.

## Impact
- **Who is affected**: Admins installing the Discord notification integration. No user impact today (latent).
- **Severity**: None currently; future risk of an unfriendly 500 on a duplicate/racing install once a DB uniqueness constraint exists.
- **Since when**: Introduced with the Discord integration in PR #275 (not yet released to prod).

## Reproduction Steps
No user-visible surface on the current code — duplicate installs are caught earlier and return a clear error, so this can't be hit through the UI. Identified by code inspection; see Technical Details for the exact path.

## Reproduction URL
https://github.com/nudgebee/nudgebee/pull/275 — `notifications-server/notifications_server/services/discord/discord.py`

---

## Technical Details
`services/discord/discord.py` does `from sqlite3 import IntegrityError`, and `save_discord_installation()` wraps its insert in `except IntegrityError: … raise WrongArgumentsException(…)`. The DB is Postgres via SQLAlchemy, which raises `sqlalchemy.exc.IntegrityError` — an unrelated class — so the `except` can never match (dead code).

Currently harmless because: (1) there's no unique constraint on `messaging_platforms (tenant_id, platform)`, so `commit()` doesn't raise; (2) the install router (`routers/tools.py` → `discord_install`) already pre-checks via `find_installation_by_tenant_and_platform(...)` and returns 400. It would bite once a unique constraint is added: a race past the pre-check would raise an **uncaught** `sqlalchemy.exc.IntegrityError` → 500.

### Fix
```python
from sqlalchemy.exc import IntegrityError
```

### Optional follow-ups (same change)
- Add a unique constraint on `messaging_platforms (tenant_id, platform)` so the guard is enforced at the DB level (and the corrected `except` becomes a real backstop).
- Roll back + close the session on the exception path — `save_discord_installation()` only closes on success, so an error leaks the connection.

Related: PR #275, feature issue #255.

shivam2003-dev:
# Description

The Discord notification integration's duplicate-install error handler can never trigger because it catches the wrong exception class. `save_discord_installation()` imports `IntegrityError` from `sqlite3`, but the DB is Postgres via SQLAlchemy, which raises `sqlalchemy.exc.IntegrityError` — an unrelated class. The `except` is dead code.

It is harmless today (duplicate installs are pre-checked in the router and there is no DB unique constraint yet), but once a uniqueness constraint is added on `messaging_platforms (tenant_id, platform)`, a racing install would raise an **uncaught** `sqlalchemy.exc.IntegrityError` and surface as a generic **500** instead of the intended clear message.

## Fix

- Import `IntegrityError` from `sqlalchemy.exc` so the handler can actually match the raised exception and return the friendly `WrongArgumentsException`.
- Roll back and close the session on the exception path. Previously `save_discord_installation()` only closed the session on the success path, so an error would leak the DB connection.

## How to verify

- `from sqlalchemy.exc import IntegrityError` is now the import used in `services/discord/discord.py`.
- On a constraint violation, `commit()` raises `sqlalchemy.exc.IntegrityError`, which the `except` now matches → `WrongArgumentsException(Err.OS0009)` (clear "already exists" message) instead of an uncaught 500.
- The session is rolled back and closed on the error path (no leaked connection).
- `python3 -m py_compile notifications-server/notifications_server/services/discord/discord.py` passes.

## Type of change

# How Has This Been Tested?

- [x] `py_compile` passes on the edited module
- [x] Verified the exception class now matches what SQLAlchemy raises for a Postgres constraint violation

Fixes #376

### DIFF
--- a/notifications-server/notifications_server/services/discord/discord.py
+++ b/notifications-server/notifications_server/services/discord/discord.py
@@ -54,10 +54,10 @@ class DiscordService:
             )
             self.session.add(installation)
             self.session.commit()
-            self.session.close()
             return installation
         except IntegrityError as exc:
             LOG.exception("Unable to save discord installation: %s", exc)
             self.session.rollback()
-            self.session.close()
             raise WrongArgumentsException(Err.OS0009, [exc])
+        finally:
+            self.session.close()
--- a/notifications-server/notifications_server/services/discord/discord.py
+++ b/notifications-server/notifications_server/services/discord/discord.py
@@ -1,6 +1,6 @@
 import logging
 import uuid
-from sqlite3 import IntegrityError
+from sqlalchemy.exc import IntegrityError
 
 from notifications_server.exceptions.common_exc import WrongArgumentsException
 from notifications_server.exceptions.exceptions import Err
@@ -58,4 +58,6 @@ class DiscordService:
             return installation
         except IntegrityError as exc:
             LOG.exception("Unable to save discord installation: %s", exc)
+            self.session.rollback()
+            self.session.close()
             raise WrongArgumentsException(Err.OS0009, [exc])


shivam2003-dev:
Good catch — addressed in the latest commit. Moved `self.session.close()` into a `finally` block so the session is released on every path (success, `IntegrityError`, or any other exception such as `OperationalError`/`DataError` during `commit()`), preventing the connection-pool leak. `close()` on an already-closed session is a safe no-op.

saiprasadpotdar:
Thanks a lot for jumping on this so fast, @shivam2003-dev  

Your fix is genuinely correct and the session handling (rollback on the error path + `close()` in `finally`) is exactly the right shape. But We're going to consolidate on #384, which makes this same import/session fix **and** adds the missing `UNIQUE (tenant_id, platform)` constraint + migration in the same change. 

So I'm closing this in favor of #384 — but that's purely a scope call, not a reflection on your patch, which was clean and ready to merge on its own.  Thanks again for the contribution.

PR #378, opened 2026-06-23
fix(ui): filter optimization trigger categories by source (#295)

### LINKED ISSUE #295: 🐛 [BUG] - Incorrect categories shown for Cloud Optimization trigger
## Description

When configuring an **Optimization Trigger** for a cloud account/cluster, the **Category** dropdown surfaces Kubernetes-specific categories (e.g. **Pod Right Sizing**, **K8s Instance Recommendation**, **K8s Spot Recommendation**) that don't apply to cloud sources, alongside the cloud-relevant ones.

## Current behavior

- Kubernetes-specific categories are shown when the trigger source is a cloud account/cluster.
- The displayed categories are not relevant to the selected source.

## Expected behavior

- Only cloud-related optimization categories should be displayed when the source is a cloud account.
- Category options should be **dynamically filtered** based on the selected optimization source/type (Cloud vs Kubernetes), so the user only sees applicable categories.

## Impact

- Users may be confused by irrelevant category options.
- Picking a K8s-only category for a Cloud trigger can lead to misconfigured automations / triggers that never fire (or fire with no matches).

## Acceptance

- Selecting a Cloud source restricts the Category dropdown to Cloud-applicable categories.
- Selecting a Kubernetes source restricts the Category dropdown to K8s-applicable categories.
- Already-saved triggers continue to render their selected category even if it's now filtered out of the dropdown (no data loss on edit).

shivam2003-dev:
# Description

When configuring an **Optimization Trigger**, the **Category** dropdown always listed every category — including Kubernetes-only ones (**Pod Right Sizing**, **K8s Instance Recommendation**, **K8s Spot Recommendation**, **K8s Missing Attribute**) — even when the selected source was a cloud account. Picking a K8s-only category for a cloud trigger produces an automation that can never match.

## Fix

In `TriggerConfigSidebar.tsx` (optimization trigger config):

- Tagged each category option with the optimization `sources` it applies to. The K8s-prefixed categories and `PodRightSizing` are `['k8s']`; `RightSizing` / `Configuration` / `Security` apply to both (`['k8s', 'cloud']`).
- Derived the selected source from the chosen cluster/account's `cloud_provider` (`K8s` → `k8s`, any cloud provider → `cloud`). The cluster options already carry `cloud_provider`; I widened the state type to expose it.
- The Category dropdown now renders only the categories applicable to the selected source. With no cluster selected, all categories are shown (unchanged).
- **No data loss on edit:** an already-saved category that is no longer in the filtered set is still appended so an existing trigger renders its selection.
- When the user actively switches the cluster to a different source, a now-inapplicable selected category is cleared so the trigger isn't left in a state that can never fire.

## How to verify

1. Create an Optimization Trigger, select a **cloud account** as the Cluster → the Category dropdown no longer shows Pod Right Sizing / K8s Instance Recommendation / K8s Spot Recommendation / K8s Missing Attribute (only Right Sizing / Configuration / Security).
2. Select a **Kubernetes cluster** → all K8s categories are available again.
3. Edit an existing trigger whose saved category is now filtered out → the saved category still renders (no data loss).
4. Switch the cluster from K8s to cloud while a K8s-only category is selected → the category resets so the trigger can't be saved in a never-matching state.
5. `tsc` reports no errors on the file.

## Type of change

# How Has This Been Tested?

- [x] TypeScript compiler reports no errors on the changed file
- [x] Manually traced the filter logic for cloud vs K8s sources and the saved-category preservation path

Fixes #295

### DIFF
--- a/app/src/components1/workflow/components/TriggerConfigSidebar.tsx
+++ b/app/src/components1/workflow/components/TriggerConfigSidebar.tsx
@@ -242,7 +242,7 @@ const TriggerConfigSidebar: React.FC<TriggerConfigSidebarProps> = ({
   const [optimizationCategories, setOptimizationCategories] = useState<string[]>([]);
   const [optimizationRuleNames, setOptimizationRuleNames] = useState<string[]>([]);
   const [optimizationClusters, setOptimizationClusters] = useState<string[]>([]);
-  const [k8sClusterOptions, setK8sClusterOptions] = useState<{ label: string; value: string }[]>([]);
+  const [k8sClusterOptions, setK8sClusterOptions] = useState<{ label: string; value: string; cloud_provider?: string }[]>([]);
   const [isLoadingK8sClusters, setIsLoadingK8sClusters] = useState(false);
 
   // Buffered trigger-config state. Edits go into `pendingTriggerConfig` instead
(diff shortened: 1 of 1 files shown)


shivam2003-dev:
Thanks for the review — applied both suggestions in the latest commit:

1. **Memoization** — `OPTIMIZATION_CATEGORY_OPTIONS`, `selectedOptimizationSource` and `filteredOptimizationCategoryOptions` are now wrapped in `useMemo` for reference stability and to avoid recomputing on every render.
2. **DRY** — extracted a reusable `getOptimizationSource(clusterValue)` helper and reused it both in the `selectedOptimizationSource` memo and the cluster `onChange` handler, removing the duplicated source-derivation logic. Left it as a plain function (no `useCallback`) since it isn't passed as a prop to a memoized child.

blue4209211 (approved):

blue4209211:
@shivam2003-dev  can you please share screenshot for this ?

app/src/components1/workflow/components/TriggerConfigSidebar.tsx
```diff
@@ -854,15 +854,52 @@ const TriggerConfigSidebar: React.FC<TriggerConfigSidebarProps> = ({
...
-    { label: 'K8s Instance Recommendation', value: 'K8sInstanceRecommendation' },
-    { label: 'K8s Spot Recommendation', value: 'K8sSpotRecommendation' },
-    { label: 'Configuration', value: 'Configuration' },
-    { label: 'Security', value: 'Security' },
-    { label: 'K8s Missing Attribute', value: 'K8sMissingAttribute' },
-  ];
+  // Each category is tagged with the optimization sources it applies to.
+  // K8s-prefixed categories (and Pod Right Sizing) are Kubernetes-only and must not
+  // show up when the selected source is a cloud account; the rest apply to both.
+  const OPTIMIZATION_CATEGORY_OPTIONS = useMemo(
```
Kankshit-02, line 860: Can we source the dropdown from recommendation_category_type table (or backend RPC) instead of hardcoding.
We can map sources on the frontend once we fetch values from recommendation_category_type.

Missing categories are InfraUpgrade and 
K8sVersionUpgrade
The deprecated ones
PodRightSizing / K8sInstanceRecommendation / K8sMissingAttribute

shivam2003-dev:
Checked the open comments here. This older PR is now superseded by #442, which carries the same optimization-trigger category fix in a smaller utility-backed implementation with tests and has also addressed the latest Gemini feedback on case-insensitive K8s source detection.

For Kankshit-02’s backend-sourced dropdown suggestion: that is a broader follow-up than this bug fix because it changes the category source of truth and needs an API/RPC contract. I would keep #442 scoped to the incorrect-category bug, then track the backend-driven category list separately if maintainers want that direction.

For the screenshot request: I will provide it on the active PR thread if maintainers still need visual confirmation there.

PR #379, opened 2026-06-23
perf(rag-server): bulk-fetch ServiceNow KB extension content (#301)

### LINKED ISSUE #301: perf(rag-server): bulk-fetch ServiceNow KB extension content to remove N+1 API calls
## Summary

The RAG ServiceNow KB scraper does one extra `GET /table/<sys_class_name>?sys_id=...` per article whose primary body is empty — this was introduced as a hotfix to recover Known Error / How-To / FAQ extended-template articles whose body lives on a child table rather than `kb_knowledge`.

On instances where extended-template articles dominate the KB this roughly **doubles the KB sync wall time** and counts against the customer's ServiceNow REST rate limit.

**Proposal:** replace the per-article fetch with a **per-template bulk fetch**: collect `sys_id`s by `sys_class_name`, then issue one `sys_idIN<csv>` query per extension table. API calls drop from O(extended articles) to O(distinct templates), which is typically 2–4.

## Where

`llm/rag-server/rag/core/documents/scraper.py`:

- `fetch_servicenow_kb_articles` (around line 360)
- `process_servicenow_kb_article` (around line 389) — currently calls
- `fetch_servicenow_kb_extension_content` (around line 312) once per empty-body article (around line 418).

## Proposed approach

After the first pass that pulls from `/table/kb_knowledge`:

1. Identify articles with empty `text` **and** empty `wiki`.
2. Group them by `sys_class_name`.
3. For each group, call `snow_client.resource("/table/<sys_class_name>").get(query="sys_idIN<csv-of-sys-ids>")` once.
4. Index the returned rows by `sys_id` and stitch into the body during `process_servicenow_kb_article`.

This replaces the current per-article `fetch_servicenow_kb_extension_content(snow_client, sys_class_name, sys_id, ...)` call on the empty-body path.

## Drawbacks / open work

- **Memory:** keeps an extra in-memory dict of extension rows during a batch. For typical KB sizes (thousands of articles, low single-digit MB per row) this is fine; very large instances may need streaming.
- **Failure blast radius:** a failure on the bulk query fails an entire template's articles instead of one — mitigated by per-article fallback on bulk failure.
- **URL length:** `sys_idIN<csv>` can grow large; chunk by ~50 `sys_id`s per call to stay under SNOW's URL limit.
- **Operator support:** ServiceNow REST `sys_idIN` operator behaviour on extension tables should be verified — some custom tables may disable IN operators.

## Open questions

- Confirm `sys_idIN` is supported on all relevant extension tables (`kb_template_known_error_article`, `kb_template_how_to_article`, `kb_template_faq_article`, plus customer customizations).
- Bulk chunk size — pick a default (~50) and make it configurable via `Config`.
- Should we add an explicit priority-map sort over the returned columns to stabilize embeddings across SNOW versions?

## Acceptance

- KB sync on a corpus dominated by extended-template articles makes ≤ O(distinct templates) extension-table calls per batch (instead of one per article).
- Behaviour-equivalent on the happy path: same article bodies recovered, same embeddings end-state.
- Per-article fallback on bulk-query failure so a misbehaving custom extension table doesn't fail the whole template.

shivam2003-dev:
# Description

The RAG ServiceNow KB scraper made one extra `GET /table/<sys_class_name>?sys_id=...` per article whose primary body extracted to nothing (extended-template articles — Known Error / How-To / FAQ — whose body lives on a child table). On instances dominated by extended-template articles this roughly doubles KB-sync wall time and counts against the customer's ServiceNow REST rate limit (an N+1 pattern).

## Fix

In `llm/rag-server/rag/core/documents/scraper.py`:

- Added `_bulk_fetch_servicenow_kb_extension_content()` which takes `{sys_class_name: [sys_id, ...]}` and issues one `sys_idIN<csv>` query per template (chunked at 50 sys_ids to stay under ServiceNow's URL-length limit), returning `{sys_id: stitched_body}`. Extension-table API calls drop from O(empty-body articles) to **O(distinct templates)** (typically 2–4).
- `collect_servicenow_kb_documents()` now does a cheap pre-pass to find articles whose primary body extracts to nothing, groups their sys_ids by `sys_class_name`, bulk-fetches once, and passes the result into `process_servicenow_kb_article()`.
- `process_servicenow_kb_article()` gained an optional `extension_content_by_sys_id` map: when the body is empty and the sys_id is present, it uses the pre-fetched content; otherwise it **falls back to the existing per-article fetch** (so a failed bulk query for one template never fails the whole sync).
- Extracted the shared `kb_`-column stitching into `_stitch_kb_extension_body()` so the per-article and bulk paths produce identical bodies (behaviour-equivalent), keeping the existing `sys_class_name` validation and plain-text `##` header safety.

## How to verify

- On a KB dominated by extended-template articles, extension-table calls are now ≤ O(distinct templates) per sync instead of one per empty-body article.
- Happy path is behaviour-equivalent: the same article bodies are recovered (both paths run the recovered content through `extract_kb_content`).
- Per-article fallback: if a template's bulk query raises, its sys_ids are omitted from the map and each such article falls back to the original single fetch.
- `python3 -m py_compile llm/rag-server/rag/core/documents/scraper.py` passes; no existing tests reference the changed functions and the public signature change is backward-compatible (new arg is optional).

## Type of change

- [x] Performance (non-breaking change which improves performance)

# How Has This Been Tested?

- [x] `py_compile` passes
- [x] Traced bulk path, per-article fallback path, and the empty-result (`kb_knowledge` / no kb_ fields) cases for behaviour-equivalence

Fixes #301

### DIFF
--- a/llm/rag-server/rag/core/documents/scraper.py
+++ b/llm/rag-server/rag/core/documents/scraper.py
@@ -335,7 +335,6 @@ def _stitch_kb_extension_body(record):
     return "\n\n".join(parts)
 
 
-
 def fetch_servicenow_kb_extension_content(snow_client, sys_class_name, sys_id, kb_number=""):
     """Fetch template-specific body fields from a KB article's extension table.
 
@@ -411,7 +410,6 @@ def _bulk_fetch_servicenow_kb_extension_content(snow_client, lookups, chunk_size
                 )
                 continue
     return result
-
 
 
 def fetch_servicenow_kb_articles(snow_client, batch_size=50):
--- a/llm/rag-server/rag/core/documents/scraper.py
+++ b/llm/rag-server/rag/core/documents/scraper.py
@@ -308,6 +308,33 @@ def get_kb_article_url(base_url, kb_number):
 # (``kb_category``, ``kb_knowledge_base``) we shouldn't embed.
 _KB_EXTENSION_SKIP_COLUMNS = frozenset({"kb_category", "kb_knowledge_base"})
 
+# Max number of sys_ids per ``sys_idIN<csv>`` bulk extension query. ServiceNow
+# has a practical URL-length limit on encoded queries, so we chunk to stay well
+# under it (sys_ids are 32 chars each).
+_KB_EXTENSION_BULK_CHUNK_SIZE = 50
+
+
+def _stitch_kb_extension_body(record):
+    """Build a plain-text body from an extension-table row's ``kb_`` columns.
+
+    OOB ServiceNow KB templates prefix their body columns with ``kb_`` (e.g.
+    ``kb_cause``, ``kb_workaround``, ``kb_question``, ``kb_answer``). Everything
+    else on the row is editor/instrumentation noise, so we whitelist the
+    ``kb_`` prefix. Reference fields come back as dicts and are skipped; only
+    non-empty string columns contribute. Uses plain-text ``##`` section headers
+    (not ``<h3>``) so an unexpected field_name can't synthesize HTML that
+    bypasses the script/style stripper in ``extract_kb_content``.
+    """
+    parts = []
+    for field_name, value in record.items():
+        if not field_name.startswith("kb_") or field_name in _KB_EXTENSION_SKIP_COLUMNS:
+            continue
+        if not isinstance(value, str) or not value.strip():
+            continue
+        parts.append(f"## {field_name}\n\n{value}")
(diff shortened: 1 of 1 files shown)


shivam2003-dev:
Checked the Gemini review. It reported no review comments or requested changes to address. Existing CI remains green.

blue4209211:
@shivam2003-dev  did we tested it with actual servicenow account ?

PR #386, opened 2026-06-24
fix(llm): stop double-counting event-analysis spend against chat budget

### LINKED ISSUE #290: [BUG] - AI chat budget runs out early because event-analysis spend is charged to it
## Description

Nudgebee tracks two separate AI spending budgets per account/organization: one for **AI chat** (user-initiated "Nubi" conversations) and one for **Event Analysis** (automatic investigation of events). These are configured as independent budgets with their own limits.

In practice the **AI chat budget is being consumed by Event Analysis spend as well** — every dollar spent on automatic event analysis is counted against the chat budget *on top of* its own event-analysis budget. As a result the chat budget shows far more usage than users have actually generated by chatting, its "remaining" figure is understated, and chat can be throttled/blocked even when users have barely used it.

## Impact

- **Who is affected**: All tenants/accounts that have AI-chat budget limits enabled and also run Event Analysis (effectively everyone using both features).
- **Severity**: AI chat can hit its budget limit and start returning "budget exceeded" errors prematurely, blocking legitimate user chats. Budget/usage reporting for chat is inflated and misleading.
- **Since when**: Unknown — present since the multi-module budget usage logic was introduced.

## Reproduction Steps

This is a backend calculation issue; it surfaces in the budget status shown for an account. To observe it:

1. Pick an account that has run Event Analysis this month (non-zero event-analysis spend).
2. Look at the budget status for that account (`ai_get_budget_status`).
3. Compare the `investigation` (Event Analysis) monthly cost usage against the `user_investigation` (AI chat) monthly cost usage.
4. Observe that the chat (`user_investigation`) usage is *always >=* the event usage and in fact includes it — chat usage minus event usage is the true chat spend.

Counts make this unmistakable: an account showing, say, 110 event conversations and 20,359 chat conversations — the 110 event conversations are a subset of the 20,359.

---

## Technical Details

Budget usage is computed per "module" in `llm/llm-server/budget`. The module → SQL filter map in `llm/llm-server/budget/types.go` is:

```go
var moduleQueryFilters = map[string]string{
    ModuleInvestigation:     " AND c.session_id LIKE '" + events.SessionIdPrefixEvent + "%'", // 'event-%'
    ModuleUserInvestigation: "", // <-- empty: matches ALL conversations, including event-*
}
```

`getEntityTokenUsage` / `getEntityConversationCount` in `llm/llm-server/budget/usage.go` apply this filter when summing cost from `llm_conversation_token_usage` (joined to `llm_model_pricing`) and counting `llm_conversations`. Because the `user_investigation` filter is empty, its usage/count query sweeps in **every** conversation for the entity — including the `event-` prefixed event-analysis sessions that `investigation` already accounts for. Hence the double-count.

This contradicts the **enforcement** logic, which is mutually exclusive by session prefix (event → `investigation`, else → `user_investigation`):

- `llm/llm-server/api/chains.go` (around lines 405, 784, 1330)
- `llm/llm-server/api/conversation_sync.go` (around line 101)

So enforcement treats the two budgets as disjoint, but usage summation treats `user_investigation` as a superset. The same `getEntityTokenUsage` backs both reporting (`GetBudgetStatus`) and enforcement (`CheckBudgetLimits`), so the inflated number also drives premature `429`s.

## Proposed fix

Make `user_investigation` the complement of `investigation`:

```go
ModuleUserInvestigation: " AND c.session_id NOT LIKE '" + events.SessionIdPrefixEvent + "%'",
```

(`event-rca-` sessions also start with `event-`, so a single `NOT LIKE 'event-%'` excludes them too.)

**Open question before fixing:** confirm that *every* non-event conversation in `llm_conversations` is genuinely an AI-chat session. If other session types exist (scheduled jobs, automations, internal agents), `NOT LIKE 'event-%'` would still lump them into chat. If so, the correct fix is a positive prefix match for chat sessions rather than a negative event match.

## Acceptance

- `user_investigation` usage no longer includes event-analysis spend.
- `GetBudgetStatus` and `CheckBudgetLimits` agree on the partition (matching the prefix-based routing in `chains.go` / `conversation_sync.go`).
- Regression test in `llm/llm-server/budget/service_test.go` asserting event spend lands only in `investigation`, not `user_investigation`.

bharathnr1:
# Description

The interactive AI chat budget runs out early because event-analysis spend is charged to it (#290).

Root cause: in `llm/llm-server/budget/types.go`, the `moduleQueryFilters` map gave `user_investigation` an **empty** filter (`""`), so its usage query matched **every** session — including `event-*` (event-analysis) sessions — and summed their spend into the chat budget. Meanwhile `investigation` correctly filters `LIKE 'event-%'`. Enforcement already attributes spend to the right module (`chains.go`/`conversation_sync.go`); only the **usage-reporting filter** was wrong, so chat usage was inflated and throttled prematurely.

Fix: make `user_investigation` the exact complement — `NOT LIKE 'event-%'` (which also covers `event-rca-*`). The two module filters are now disjoint.

Fixes #290

## Type of change

# How Has This Been Tested?

- [x] `go build ./budget/`, `gofmt` — clean
- [x] Added `TestModuleQueryFiltersComplementary` (deterministic, no DB) asserting `investigation` uses `LIKE 'event-%'`, `user_investigation` uses `NOT LIKE 'event-%'`, and the latter is non-empty — a direct regression guard against reverting to the empty filter. Passes.

- **Complement correctness:** `event-rca-*` sessions start with `event-`, so they match `investigation`'s `LIKE 'event-%'` and are excluded from `user_investigation` by the same prefix — no session falls in both or neither.
- **Test scope:** the budget package has no sqlmock harness (existing tests are pure unit tests), so the regression test asserts the filter strings encode the complement relationship rather than exercising a live query. It still fails loudly if anyone reverts `user_investigation` to an empty/over-broad filter, which is the actual bug.

### DIFF
--- a/llm/llm-server/budget/service_test.go
+++ b/llm/llm-server/budget/service_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"nudgebee/llm/config"
+	"nudgebee/llm/events"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -52,6 +53,24 @@ func TestValidModules(t *testing.T) {
 		assert.True(t, validModules[module], fmt.Sprintf("%s should be a valid module", module))
 	}
 	assert.GreaterOrEqual(t, len(validModules), len(expectedModules))
+}
+
+// TestModuleQueryFiltersComplementary guards against regression of #290: the
+// user_investigation filter was empty, so it matched every session (including
+// event-*), double-counting event-analysis spend against the interactive chat
+// budget. The two module filters must be complements over the event- prefix.
+func TestModuleQueryFiltersComplementary(t *testing.T) {
+	inv := moduleQueryFilters[ModuleInvestigation]
+	user := moduleQueryFilters[ModuleUserInvestigation]
+
+	// investigation counts only event-* sessions.
+	assert.Contains(t, inv, "LIKE '"+events.SessionIdPrefixEvent+"%'")
+	assert.NotContains(t, inv, "NOT LIKE")
+
+	// user_investigation must be the complement — and must NOT be empty (an
+	// empty filter was the bug: it swept in event-* sessions).
+	assert.NotEmpty(t, user, "user_investigation filter must exclude event-* sessions, not be empty")
+	assert.Contains(t, user, "NOT LIKE '"+events.SessionIdPrefixEvent+"%'")
 }
 
 // TestValidEntityTypes tests the entity type validation
--- a/llm/llm-server/budget/types.go
+++ b/llm/llm-server/budget/types.go
@@ -20,8 +20,11 @@ var validModules = map[string]bool{
 // moduleQueryFilters maps module names to their SQL query filters
 // This is a whitelist approach to prevent SQL injection
 var moduleQueryFilters = map[string]string{
-	ModuleInvestigation:     " AND c.session_id LIKE '" + events.SessionIdPrefixEvent + "%'",
-	ModuleUserInvestigation: "", // No additional filter for user investigation
+	ModuleInvestigation: " AND c.session_id LIKE '" + events.SessionIdPrefixEvent + "%'",
+	// user_investigation is the complement of investigation: exclude event-*
+	// sessions (incl. event-rca-*) so event-analysis spend isn't double-counted
+	// against the interactive chat budget.
+	ModuleUserInvestigation: " AND c.session_id NOT LIKE '" + events.SessionIdPrefixEvent + "%'",
 }
 
 // Entity type constants


bharathnr1:
No blocking notes from the bot here — the `NOT LIKE` complement plus the regression test cover it. Rebased on latest main.

hsundar-git:
Thanks a lot for this, @bharathnr1 — the fix is correct and the analysis is spot-on. 🙏

Heads-up though: there's an earlier PR for the exact same change — #370 (by @PoorneshThatturi), opened on Jun 23. Both PRs make the identical `user_investigation` → `NOT LIKE 'event-%'` partition fix in `budget/types.go`. Since #370 came first (and carries the same fix with sqlmock-backed regression tests), we'll consolidate on that one.

Closing this as a duplicate of #370 — no reflection on the quality of your work, which was great.

For next time: before starting on a task, could you do a quick scan of the open PRs/issues for that area first? It helps us avoid two people landing the same fix in parallel. Really appreciate the contribution and looking forward to the next one! 🚀

Closing in favor of #370.

PR #389, opened 2026-06-24
fix(llm): batch ServiceNow KB extension-table fetches to kill N+1

### LINKED ISSUE #301: perf(rag-server): bulk-fetch ServiceNow KB extension content to remove N+1 API calls

bharathnr1:
# Description

When syncing ServiceNow knowledge-base articles, `collect_servicenow_kb_documents` fetched the batch of articles in one call, but then issued **one GET per article** whose primary body was empty — the extended templates (known-error / how-to / FAQ) that store their body on extension tables `/table/kb_knowledge` doesn't project. On a KB dominated by those, this roughly doubles API calls to **O(N)**: slow sync wall-time and rate-limit pressure on the customer's instance.

This batches the extension-table fetches: empty-bodied articles are grouped by `sys_class_name` and resolved with **one `sys_idIN` query per class** (chunked at 100) instead of one per article.

Fixes #301

## Key points

- `fetch_servicenow_kb_extension_content_bulk` (new) uses pysnow `QueryBuilder().field("sys_id").equals([ids])`, which emits `sys_idINa,b,c` for list values (`equals` switches to `IN`; pysnow 0.7.17 has no `is_one_of`). The single-article `fetch_servicenow_kb_extension_content` survives as a fallback for when the bulk pass skipped an article or its chunk raised.
- Attempted-but-empty `sys_id`s are seeded into the result map, so genuinely-empty articles are dropped **without** a redundant single fetch; only a bulk-query *exception* leaves an id unseeded → per-article fallback.
- Extracted `_is_valid_extension_class` + `_extract_kb_extension_fields` so the single and bulk paths share validation (table-name regex guard) and `kb_`-prefix field extraction — no drift.
- Pure-local change to `scraper.py`: no DB migration, no cross-service contract, no guarded areas.

## Type of change

- [x] Performance (non-breaking change which improves performance)

# How Has This Been Tested?

New `tests/test_servicenow_scraper.py` (6 tests) with a fake pysnow client and the real `QueryBuilder` — no live ServiceNow:

- [x] One bulk query per class (not per article); bodies resolve correctly
- [x] Normal-bodied articles never touch the extension table
- [x] Genuinely-empty article dropped with no redundant single fetch
- [x] Bulk-query failure falls back to per-article single fetch
- [x] Large id lists chunk into multiple bulk queries
- [x] Malformed `sys_class_name` rejected before any API call
- [x] `black --check` and `flake8` (incl. `max-complexity=18`) clean

- **Extension rows share the base article's `sys_id`** (ServiceNow table inheritance), so indexing bulk results by `sys_id` and matching them back to articles is correct. If a customer instance violated that, the body would simply not resolve and fall through to the existing "no content" warning — no crash, no wrong attribution.
- **Primary-body emptiness is computed twice** (once in `_bulk_fetch_extension_content` to decide grouping, once in `process_servicenow_kb_article`) — pure CPU (BeautifulSoup), not API; kept for a minimal, readable diff rather than threading precomputed bodies through the signature.
- **Chunk size 100** bounds `sys_idIN` URL length; ServiceNow has no documented hard cap but 100 is comfortably safe and keeps each call cheap.

### DIFF
--- a/llm/rag-server/rag/core/documents/scraper.py
+++ b/llm/rag-server/rag/core/documents/scraper.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 
 import pysnow  # type: ignore[import-untyped]
 import requests
+from pysnow import QueryBuilder  # type: ignore[import-untyped]
 from atlassian import Confluence
 from bs4 import BeautifulSoup, NavigableString
 from filelock import FileLock, Timeout
@@ -309,8 +310,27 @@ def get_kb_article_url(base_url, kb_number):
 _KB_EXTENSION_SKIP_COLUMNS = frozenset({"kb_category", "kb_knowledge_base"})
 
 
-def fetch_servicenow_kb_extension_content(snow_client, sys_class_name, sys_id, kb_number=""):
-    """Fetch template-specific body fields from a KB article's extension table.
+def _is_valid_extension_class(sys_class_name, kb_number=""):
+    """Whether ``sys_class_name`` names a KB *extension* table worth querying.
+
+    ``kb_knowledge`` is the base table (body already projected) and the empty
+    string means "no subclass", so neither needs an extension fetch. The value
+    is interpolated into the API path, so we also reject anything that doesn't
+    match a ServiceNow table-name shape (defense-in-depth against a
+    misconfigured / hostile upstream returning unexpected values).
+    """
+    if not sys_class_name or sys_class_name == "kb_knowledge":
+        return False
+    if not re.fullmatch(r"[a-z][a-z0-9_]*", sys_class_name):
+        logger.warning(
+            f"Refusing to fetch extension content for {kb_number}: " f"unexpected sys_class_name={sys_class_name!r}"
+        )
+        return False
+    return True
+
+
+def _extract_kb_extension_fields(record):
+    """Concatenate the ``kb_``-prefixed body columns of an extension-table row.
 
     OOB ServiceNow KB templates (``kb_template_known_error_article``,
     ``kb_template_how_to_article``, ``kb_template_faq_article``, ...) all prefix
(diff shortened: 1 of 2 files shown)


bharathnr1:
Closing as a duplicate of #379, which addresses the same ServiceNow KB extension-table N+1 (#301) and was opened first. The two PRs use essentially the same design (bulk `sys_idIN` per template, optional pre-fetch map, per-article fallback), so #379 should be the canonical fix.

One thing this PR added that #379 doesn't have is a unit-test suite (`tests/test_servicenow_scraper.py`, 6 tests with a fake pysnow client covering: one bulk query per class, no redundant single fetch for genuinely-empty articles, per-article fallback on bulk failure, chunking, and malformed `sys_class_name` rejection). Happy to contribute those as a follow-up PR on top of #379 if the maintainers find them useful. cc @shivam2003-dev

PR #392, opened 2026-06-24
feat(splunk): discover log fields dynamically instead of hardcoded list

### LINKED ISSUE #380: Splunk Observability Cloud: discover log fields dynamically instead of a hardcoded list
## Summary

Nudgebee's Splunk Observability Cloud (formerly SignalFx) log source advertises log **field names** from a hardcoded static list rather than discovering them from the backend. Any custom OpenTelemetry attribute a user indexes in Splunk Log Observer is therefore missing from field autocomplete and label-value lookups.

## Where

`api-server/services/observability/splunk_logs.go`

`SplunkLogSource.QueryLabels` returns a fixed slice of well-known OTel field names, with an explicit in-code acknowledgement:

```go
// A full dynamic implementation would call a catalog API; for now return the standard set.
knownFields := []string{
    "message", "severity", "timestamp",
    "kubernetes.namespace.name", "kubernetes.pod.name", ...
}
```

`QueryLabelValues` is also approximate — it samples ~500 recent log entries and extracts distinct values rather than asking the backend for the field's value set.

## Why it matters

- Field autocomplete in the log query builder only ever shows the ~11 baked-in fields. Custom fields (e.g. `service.version`, `http.status_code`, app-specific attributes) never appear.
- Behaviour is inconsistent with other providers (Elasticsearch, Loki, Dynatrace) whose `QueryLabels` reflect what is actually present in the data.

## Proposed change

1. Implement `QueryLabels` to enumerate available log fields from Splunk Log Observer (e.g. by deriving the field set from a recent-results sample over the requested time range, or a catalog/fields API if one is available for the Log Observer dataset).
2. Keep the current static list as a fallback when the dynamic call returns nothing or errors, so the field list never regresses to empty.
3. (Optional) Improve `QueryLabelValues` to honour the requested time range and field consistently.

## Acceptance criteria

- [ ] A custom OTel log attribute present in Splunk Log Observer appears in `QueryLabels` output.
- [ ] When dynamic discovery fails, the existing well-known field set is still returned (no empty/error regression).
- [ ] Unit test covering: dynamic fields surfaced, and fallback path on error.

## Good first issue notes

The change is self-contained to one file and mirrors patterns already used by other log sources in the same package. Look at how `QueryLabelValues` already samples recent entries for a starting point.

VaishnaviKumar0704:
Closes #380

## Summary
Replaces the hardcoded field list in `SplunkLogSource.QueryLabels` with 
dynamic discovery by sampling recent log entries, so custom OTel attributes 
(e.g. `service.version`, `http.status_code`, app-specific fields) appear in 
field autocomplete automatically.

## Changes
- `QueryLabels` now samples up to 200 recent log entries and collects all 
  attribute keys dynamically
- The static well-known field list is kept as a fallback — if the dynamic 
  call fails or returns nothing, the field list never regresses to empty
- Extracted `splunkO11yKnownFields` to a package-level var shared by both 
  production code and tests

## Tests
Added `splunk_logs_test.go` covering:
- Dynamic fields from sample entries are surfaced in output
- Static fallback is returned on search error (no empty list, no error)
- Static fallback is returned when sample returns zero entries


mayankpande88:
Closing as a duplicate of #390, which was just merged for #380. #390 takes the same dynamic-discovery approach but also merges discovered fields with the static fallback set (so standard fields never drop on a sparse sample) and ships unit tests for the fallback paths. Thanks for the work here, the two were very close.

PR #406, opened 2026-06-24
fix(collector): return AWS config error in getAwsUsageReport instead of nil (#405)

### LINKED ISSUE #405: bug(collector): AWS usage report swallows config error, returning empty success
## Description

`getAwsUsageReport` (`collector-server/cloud-collector/providers/aws/usage_report.go`) returns a **nil error** when AWS config resolution fails:

```go
cfg, err := getAwsConfigFromAccount(ctx.GetContext(), account)
if err != nil {
    return providers.GetUsageReportResponse{}, nil   // <- swallows err
}
```

So a broken/unauthorized AWS account is reported to the caller as a successful run with an empty usage report — the cost-data collection silently produces no data instead of surfacing the failure.

## Fix

Return the wrapped error instead of nil.

bharathnr1:
# Description

`getAwsUsageReport` returned `(GetUsageReportResponse{}, nil)` when `getAwsConfigFromAccount` failed, so a broken/unauthorized AWS account was reported to the caller as a successful run with an empty usage report — silent cost-data loss. Now returns the wrapped error.

Fixes #405

## Type of change

# How Has This Been Tested?

- [x] `go build ./providers/aws/` clean; gofmt clean. One-line error-propagation fix.

### DIFF
--- a/collector-server/cloud-collector/providers/aws/usage_report.go
+++ b/collector-server/cloud-collector/providers/aws/usage_report.go
@@ -513,7 +513,7 @@ func parseBillingPeriodStart(commonPrefix, pathPrefix string) (time.Time, bool) 
 func getAwsUsageReport(ctx providers.CloudProviderContext, account providers.Account, month time.Month, year int) (providers.GetUsageReportResponse, error) {
 	cfg, err := getAwsConfigFromAccount(ctx.GetContext(), account)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mayankpande88 (approved):
Reviewed in depth and verified against the relevant interfaces, DB constraints, and call sites. CI is green and the bot suggestions are already addressed in the diff. LGTM.

PR #414, opened 2026-06-24
fix(llm-server): fix goroutine leak on timeout in K8s resource search

### LINKED ISSUE #172: fix(llm-server): goroutine leak on timeout in K8s resource search (`tool_resource_search.go`)
## Summary

`SearchResources` in `llm/llm-server/tools/tool_resource_search.go` runs four parallel search goroutines and aggregates results in a `select` loop with a timeout. The result channels are **unbuffered**. On timeout, the loop breaks and stops receiving — but the still-running goroutines will eventually try to send on their channel and **block forever**. Each timed-out search permanently leaks up to four goroutines (and the data they captured).

Under repeated slow searches this grows unbounded.

## Where

`llm/llm-server/tools/tool_resource_search.go`:
- Unbuffered channel declarations: lines **822–825**
- Timeout-and-break path: lines **884–909**

```go
// Channels to collect results from parallel execution
commonResChan := make(chan searchResult)   // unbuffered
clusterResChan := make(chan searchResult)  // unbuffered
crdResChan := make(chan searchResult)      // unbuffered
labelResChan := make(chan searchResult)    // unbuffered

// ... 4 goroutines, each ending with `someChan <- searchResult{...}` ...

CollectResults:
for i := 0; i < 4; i++ {
    select {
    case res := <-commonResChan: /* ... */
    case res := <-clusterResChan: /* ... */
    case res := <-crdResChan: /* ... */
    case res := <-labelResChan: /* ... */
    case <-timeout:
        nbRequestContext.Ctx.GetLogger().Warn("resource-search: timeout waiting for strategies to complete")
        break CollectResults  // <-- stops receiving; senders block forever
    }
}
```

## Why it's a bug

The four goroutines have no way to know the loop stopped reading. They each call `someChan <- result` on completion. Without a receiver and without a buffered slot, the send blocks indefinitely → goroutine never exits → memory captured by the goroutine never gets reclaimed.

## Suggested fix (simple)

Give each channel capacity 1 so a late send never blocks:

```go
commonResChan := make(chan searchResult, 1)
clusterResChan := make(chan searchResult, 1)
crdResChan := make(chan searchResult, 1)
labelResChan := make(chan searchResult, 1)
```

This is the minimal fix and correct because each goroutine sends exactly once.

## Suggested fix (proper)

Thread a cancellable context into each search goroutine and have them check `ctx.Done()` so they exit cleanly on timeout rather than running their full search anyway:

```go
ctx, cancel := context.WithTimeout(nbRequestContext.Ctx, timeoutDuration)
defer cancel()

go func() {
    select {
    case commonResChan <- runCommonSearch(ctx, ...):
    case <-ctx.Done():
    }
}()
```

This avoids the leak *and* stops doing work the caller no longer wants.

## Why help-wanted (not good-first-issue)

Requires concurrency reasoning to understand why the unbuffered channel + `break` combination leaks. The buffered-channel fix is mechanical; the context-based fix is the right architectural answer.

## Validation

After the fix, write a test that:
1. Launches `SearchResources` with a short timeout.
2. Forces the search goroutines to take longer than the timeout (mock the underlying K8s calls).
3. Asserts `runtime.NumGoroutine()` returns to its baseline within a few seconds after the call returns.

## Suggested label

Once accepted, also consider `priority:high` if production logs show goroutine count growing over time.

chaitanya867:
Fixes #172

SearchResources launches 4 goroutines that each send on their respective channel once done. The channels were unbuffered, so any goroutine that finished after the collect loop timed out would block forever on the send with no receiver alive — causing a permanent goroutine and memory leak that grows under repeated slow searches.

Fix: buffer each channel (cap 1) so a late-finishing goroutine can always send and exit. Additionally propagate a cancellable context derived from the request context so goroutines exit their inner resource-type loops promptly when the caller times out, rather than continuing unnecessary kubectl calls.

Adds a TestFindActualResources_NoGoroutineLeak regression test using runtime.NumGoroutine() that runs 5 back-to-back calls and verifies the goroutine count does not accumulate (a leaked 4 goroutines-per-call pattern would add 20 goroutines, well above the +5 tolerance).


mayankpande88 (changes requested):
I dug into this one against the head file and it has two fundamental problems.

**1. It does not compile.** The goroutines reference `ctx.Done()` (lines ~840, 862, 884, 909) and the collect loop calls `cancel()` (line ~950), but `ctx` and `cancel` are never declared anywhere in the function or file. There is no `context.WithCancel(...)` / `context.WithTimeout(...)`, I grepped the whole file. So this fails to build three ways: `undefined: ctx`, `undefined: cancel`, and `imported and not used: context`. It looks like the consumer half of a cancellation mechanism was added without the producer half (`ctx, cancel := context.WithCancel(nbRequestContext.Ctx.GetContext())`).

**2. The bug being fixed does not exist as described.** The PR says the channels were unbuffered so a late goroutine blocks forever on send. But in `main` the channels are already buffered at cap 1 (`make(chan searchResult, 1)`), and the diff does not change those lines. Each goroutine sends exactly once on its own cap-1 channel, so the send can never block even with no receiver. The described leak is not possible in the current code, which also means `TestFindActualResources_NoGoroutineLeak` would pass on plain `main` and cannot catch a regression.

If issue #172 describes a real leak, it is somewhere else (likely inside `searchResourceType` or `expandWorkloadResources`), not the channel sends. The one good idea here, propagating a cancel context to stop wasted kubectl calls after timeout, is worth doing but needs the context actually created and wired, and the `ctx.Done()` checks only fire between resource-type iterations so they will not abort an in-flight call anyway. Suggest closing or reworking.

blue4209211:
@chaitanya867  @DLTKChaitanyaChougule  can you please sign CLA
https://github.com/nudgebee/nudgebee/pull/414#issuecomment-4791773504

looks like 2 seperate Ids are used.. because of that its failing

PR #415, opened 2026-06-24
feat(workflow): add Blast Radius Pre-Flight workflow node

chaitanya867:
Adds `core.blast-radius-preflight` — a new workflow task that queries the Knowledge Graph to enumerate all downstream services reachable from a target resource (up to 3 hops) before any destructive action executes.

## What it does

Before a pod restart, scale-down, failover, or config change runs, this node:

1. Resolves the target resource name → Knowledge Graph node IDs via `kg_list_nodes`.
2. Traverses downstream dependencies (direction: downstream, max_depth: 3) via `kg_list_path`.
3. Computes a weighted risk score (0–100):
   - +2 per downstream node (capped at 50)
   - +10 per "heavy" node kind — database, Redis, RabbitMQ, Kafka, external services (capped at 50)
4. Maps the score to a risk level: low / medium / high / critical.
5. Optionally calls the LLM (include_llm_verdict=true) to generate a human-readable VERDICT and RECOMMENDATION.
6. When on_high_risk=block and risk_level >= risk_threshold: returns an error that halts the workflow. When on_high_risk=warn: proceeds and surfaces the risk report in task output.

## Files changed

- `runbook-server/internal/tasks/core/blast_radius_preflight_task.go` New task implementation (~270 lines). Pure Go, no new dependencies. Uses existing `common.HttpPost`, `config.Config.ServiceEndpoint`, and `llm.ProcessRequest` — same patterns as all other tasks in this package.

- `runbook-server/internal/tasks/core/blast_radius_preflight_task_test.go` 19 unit tests covering: risk scoring, score/level thresholds, capping invariants, LLM response parsing, schema validation, riskOrder monotonicity. All tests are pure-logic (no live cluster or LLM required).

- `runbook-server/internal/tasks/registry.go` Registers `&core.BlastRadiusPreflightTask{}` alongside `ApprovalTask`.

## Example workflow usage

```yaml
tasks:
  - id: preflight
    type: core.blast-radius-preflight
    params:
      resource_name: "{{ Inputs.target_service }}"
      namespace: production
      action_description: "scaling down to 1 replica for maintenance"
      risk_threshold: high
      on_high_risk: block
      include_llm_verdict: true

  - id: scale_down
    type: k8s.kubectl
    depends_on: [preflight]
    params:
      command: "kubectl scale deployment/{{ Inputs.target_service }} --replicas=1 -n production"
```


mayankpande88 (changes requested):
Solid structure and good test coverage. Two things to resolve before merge, one mechanical and one about correctness.

**1. gofmt fails (CI blocker).** Running `gofmt -d` on `blast_radius_preflight_task.go` shows the `Execute` function has misaligned short declarations (`namespace   := ...`) and an over-padded `headers` map literal. golangci-lint will reject this. `make fmt` fixes it.

**2. The traversal direction looks inverted.** The KG defines (per `actions.yaml` and the SQL in `knowledge_graph/core/service.go`): `downstream` = "what X depends on" (follows source -> destination), `upstream` = "what depends on X". The blast radius of touching X, meaning the services that break when you restart or scale it down, is the set that depends on X, i.e. `upstream`. This task queries `direction: "downstream"`, which enumerates X's own dependencies instead. So restarting `checkout-service` would report the database checkout talks to rather than the services that call checkout. Unless edge directionality here is modeled opposite to the documented convention, this measures the wrong set. Could you confirm against a real graph?

**Design point worth addressing:** the gate fails open silently. If `kg_list_nodes`/`kg_list_path` errors (KG down, not built, auth fails), the code logs a warning, proceeds with zero nodes, computes `risk_score=0`, and `is_blocked=false`. A gate whose job is to halt destructive actions then quietly disables itself on any error. For `on_high_risk=block` consider failing closed, or returning a distinct "assessment unavailable" state rather than "low risk".

Minor: `KgNodeSlim` has no `namespace` field, so `affected_nodes[].namespace` is always empty despite the output schema advertising it. And the auth header name is hardcoded `"X-ACTION-TOKEN"` instead of `config.Config.ServiceApiServerTokenHeader` (works on the default, breaks if a deployment customizes it).

PR #416, opened 2026-06-24
feat(task-runner): add standalone Task Runner page under Automations

### LINKED ISSUE #296: [REQUEST] - Add a standalone Task Runner page to the frontend
## Summary

The backend already exposes a `POST /tasks/:task_type/execute` endpoint (RPC action `workflow_trigger_task`) for executing individual tasks outside of a workflow context. On the frontend, `TRIGGER_TASK` is already wired through the API layer and used for dry-run testing inside the workflow builder.

What's missing is a **standalone Task Runner page** — a place where a user can pick any task type, fill its parameters, and run it directly without first authoring a full workflow.

## Proposal

- Add a **Task Runner** page reachable from the main navigation or the Workflows section.
- List all available task types, grouped by category (HTTP, Kubernetes, AWS, Notification, etc.).
- Selecting a task type renders its parameter form (reuse the existing task config components from the workflow builder).
- A **Run** button calls `workflow_trigger_task` with the configured parameters.
- Render the task execution result (output, logs, errors) in a results panel below the form.
- Optional: allow saving frequently used task configurations as "quick actions".

## Drawbacks / open work

- Reusing workflow-builder task forms outside the workflow context may require some refactoring to decouple them from builder-only state.
- Tasks executed outside workflows have no execution history tracking unless the backend grows support for it (see open questions).

## Open questions

- Should standalone task executions be logged and visible in execution history?
- Should there be permission controls on which task types can be executed standalone?
- Should task configurations be saveable / shareable across users?

Anupkankale:
## Summary
Closes #296

Adds a standalone Task Runner page nested under the Automations section.

image

## Changes
- `app/src/components1/workflow/TaskRunner.tsx` — new component with task listing grouped by category, dynamic param form from `input_schema`, run button, and results panel
- `app/src/pages/auto-pilot/task-runner/index.tsx` — new page route with `tenant_admin` / `account_admin` RBAC gate
- `app/src/pages/auto-pilot/index.jsx` — added Task Runner tab (value: 2, fragment: task-runner)

image

## What's included in this PR
- [x] Task type listing grouped by category with search
- [x] Dynamic param form built from task `input_schema`
- [x] Run button wired to `workflow_execute_task` via `apiWorkflow.triggerTask()`
- [x] Results panel showing JSON output
- [x] RBAC gate for `tenant_admin` and `account_admin` only

## Out of scope (per maintainer guidance)
- Execution history logging  separate backend ticket
- Saving task configurations  future enhancement


mayankpande88 (dismissed):
Thanks for this. One blocker before it can go in:

**Runtime crash from a missing import.** In `app/src/pages/auto-pilot/index.jsx` the new tab uses `getUserSession()?.roles`, but the file never imports `getUserSession` (it only imports `hasWriteAccess` from `@lib/auth`, and this PR adds only the `TaskRunner` import). That is a `ReferenceError` thrown while building `filterOptions`, so it crashes the whole AutoPilot page render, not just the new tab. The standalone `task-runner/index.tsx` imports it correctly, so the fix is just adding `getUserSession` to the `@lib/auth` import in `index.jsx`.

A few non-blocking things worth addressing too:

- RBAC here is client-side only (both the tab and the page gate on `getUserSession().roles`). That hides the UI but does not stop anyone from calling `workflow_execute_task` directly. Please confirm the action is gated server-side as well.
- `params` is built as `Record<string, string>` from the text fields, so tasks whose `input_schema` expects integers/booleans/objects will receive strings. Consider coercing against `schema.type`.
- There are now two ways to reach this feature: the tab inside `/auto-pilot` (value 2) and the standalone `/auto-pilot/task-runner` page. Pick one so they do not drift.
- Title still reads `[Draft v2] Feachure` (typo).

The API wiring (`listTaskDefinitions`, `triggerTask`, and the `workflow_list_taskdefinitions` / `workflow_execute_task` response shapes) all check out.

Anupkankale:
Thanks for the thorough review @mayankpande88!

1. Fixed missing getUserSession import in index.jsx
   (import { hasWriteAccess, getUserSession } from '@lib/auth')

2. Removed standalone task-runner/index.tsx — tab in /auto-pilot 
   is the single entry point now

3. Server-side gate confirmed — workflow_execute_task in actions.yaml 
   has permissions for both tenant_admin and account_admin, so backend 
   rejects unauthorized calls even if frontend is bypassed

4. Type coercion (string → int/bool) — will address in a follow-up PR

5. Fixed PR title typo

blue4209211:
Thanks for this — the API wiring is solid (`listTaskDefinitions` / `triggerTask` / the `workflow_*` shapes all line up). A few things to address before this can merge (skipping the CI lint errors since they're already visible in the checks):

### 1. The `tenant_admin` / `account_admin` RBAC gate is missing (please prioritize)
The checklist and description claim an admin-only RBAC gate (and a `pages/auto-pilot/task-runner/index.tsx` route), but neither is in the diff — the change is just a tab in `auto-pilot/index.jsx` with no role check. The unused `getUserSession` import looks like the gate was intended but never wired up. As written, the Task Runner tab **and its privileged `workflow_execute_task` mutation are exposed to every user, including read-only/non-admins.** Please gate both the tab visibility and the render on the admin role before merge — this is the main blocker.

### 2. `getCategory` grouping is broken (`TaskRunner.tsx`)
```js
if (name.includes('')) return name.split('')[0].toUpperCase();
```
`name.includes('')` is always `true`, and `name.split('')[0]` returns the **first character** — so every task ends up grouped under its first letter and the camelCase fallback below is dead code. Looks like a `'_'` was lost; did you mean:
```js
if (name.includes('_')) return name.split('_')[0].toUpperCase();
```
Right now the headline "grouped by category" feature doesn't actually group by category.

### 3. Missing newline at end of `TaskRunner.tsx`
`prettier:check` will fail on this once the lint errors are cleared. `npm run lint2:fix` handles it.

### 4. Description vs. diff
The PR body references a standalone page route and an RBAC gate that aren't present. Please update it to match what actually ships (a tab inside `auto-pilot`) so reviewers aren't working from the wrong mental model.

### Minor
- **No NaN guard in type coercion:** `parseInt`/`parseFloat` on an invalid `integer`/`number` field silently sends `NaN` to the backend — worth validating or falling back.
- Coercion only handles scalar `type`s (integer/boolean/number/string); array/object/enum params fall through as raw strings. Fine for v1, just flagging.

(1) and (2) are the substantive ones; the rest are quick. Happy to clarify anything here.

---
_Generated by [Claude Code](https://claude.ai/code)_

Anupkankale:
Thanks for the detailed review @blue4209211 

Here's what's been addressed in a7de2df:

1. ✅ **RBAC gate added** — `getUserSession()` is now wired up. `isAdmin` checks for `tenant_admin` or `account_admin` in `session.roles`. The Task Runner tab is disabled (`disabled: !isAdmin`) and the render is blocked (`selectedFilter === 2 && isAdmin &&`) for non-admin users.

2. ✅ **`getCategory` fixed** — changed `split('')` to `split('_')` so tasks now correctly group by underscore-separated prefix instead of first character.

3. ✅ **Prettier fixed** — `npm run lint2` passes clean with zero errors and zero warnings.

4. ✅ **NaN guard added** — `parseInt` and `parseFloat` results are now checked with `isNaN()` and fall back to the raw string value if invalid.

Regarding point 4 (description vs diff) — I'll update the PR body now to accurately reflect that this ships as a tab inside the existing auto-pilot page, not a standalone route.

Let me know if anything else needs attention!

Anupkankale:
"Hi!  @blue4209211  Thanks for the thorough review. I have addressed all the feedback:

  **1. Admin RBAC Gate:**
      • Both the tab visibility and tab content rendering are now
      strictly gated on the admin role. Non-admin users will not see the
      Task Runner tab in the UI, and the URL hash router is configured to
      ignore  #task-runner for non-admin accounts.

  **2. Category Grouping:**
      • Verified that  getCategory  in  TaskRunner.tsx  correctly checks
      name.includes('_') . The accordion categories group tasks by their
      prefixes (e.g.  K8S_DEPLOY  is grouped under  K8S ) and
      successfully fall back to parsing camelCase formats.

  **3. Prettier Formatting:**
      • Ran  npm run lint2:fix  to ensure all files conform to prettier
      checks, including the ending newline in  TaskRunner.tsx .

  **4. Parameter Coercion & NaN Guard:**
      • Added robust  isNaN  checks for  integer  and  number  types.
      Invalid numeric entries will now throw a descriptive validation
      error (e.g.  Parameter "xyz" must be an integer. ) that displays as
      a critical banner on the frontend.
      • Optional parameters left empty are skipped (preventing empty
      string payloads from being sent to Go backend integer/number
      endpoints).
image

  **5. PR Description Updated:**

      • Updated the PR summary to clarify that this ships as a scoped tab
      in  /auto-pilot  rather than a standalone page route."***

app/src/components/workflow/TaskRunner.tsx
```diff
@@ -0,0 +1,296 @@
+import React, { useEffect, useState } from 'react';
+import { Box, Typography, CircularProgress, Accordion, AccordionSummary, AccordionDetails } from '@mui/material';
```
Kankshit-02: Can we use existing components from the ds.
Also for the icons

Kankshit-02 (commented):
Please add screenshots after updating with ds

Kankshit-02 (approved):

PR #435, opened 2026-06-26
fix: NB-434 close response body on 401/500 error paths in services_server

### LINKED ISSUE #434: 🐛 [BUG] -  HTTP Response-Body / Connection Leak on Error Paths in services_server
### Pre-flight checklist

### Affected component(s)

services-server, llm-server

### Version / commit

_No response_

### Description

### Expected behavior

### Actual behavior

### Reproduction steps

1. Open **service.go**.

### Frequency

None

### Reproduction URL

_No response_

### Screenshots

_No response_

### Logs

```bash

```

### Browsers (UI only)

_No response_

### OS

Windows

mohitkumhar:
This PR fixes a critical file descriptor and TCP connection leak in `llm/llm-server/services_server/service.go`. 

Previously, in `ExecuteQuery`, `ExecuteScanImageQuery`, and `ExecuteScanCisQuery`, the `defer resp.Body.Close()` statement was placed *after* the early returns for `401` and `500` status codes. Because of this, if the upstream server returned an error, the response body was never closed, leaking the connection. 

This fix simply moves the `defer` block up to sit immediately after the `if err != nil` check, ensuring the response body is always closed regardless of the HTTP status code returned.

Fixes #434

### DIFF
--- a/llm/llm-server/services_server/service.go
+++ b/llm/llm-server/services_server/service.go
@@ -31,6 +31,12 @@ func ExecuteQuery(serviceRequest ServicesQueryRequest) (map[string]string, error
 		return response, fmt.Errorf("services: executequery, unable to process request: %v", err)
 	}
 
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			slog.Info("services_server: failed to close response body", "error", err)
+		}
+	}()
+
 	if resp.StatusCode == 401 {
 		return response, fmt.Errorf("unauthorized: %v", resp.Body)
 	}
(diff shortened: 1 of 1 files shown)


RamanKharchee (approved):

RamanKharchee:
@mohitkumhar Please complete the CLA process. https://github.com/nudgebee/nudgebee/pull/435#issuecomment-4807266371

mohitkumhar:
> @mohitkumhar Please complete the CLA process. [#435 (comment)](https://github.com/nudgebee/nudgebee/pull/435#issuecomment-4807266371)

Hey,
Done!! I have completed the CLA signing process

blue4209211:
@mohitkumhar  thanks for fixing the issue..
this is already fixed as part of https://github.com/nudgebee/nudgebee/pull/445

so I am closing this PR

PR #437, opened 2026-06-26
fix(llm): close services response bodies on error paths

### LINKED ISSUE #434: 🐛 [BUG] -  HTTP Response-Body / Connection Leak on Error Paths in services_server

shivam2003-dev:
# Description

Fixes #434

`services_server.ExecuteQuery`, `ExecuteScanImageQuery`, and `ExecuteScanCisQuery` returned immediately on services-server `401`/`500` responses before closing `resp.Body`. During repeated upstream auth failures or 500s, those paths could leak response bodies and keep HTTP connections/file descriptors open.

This PR moves body cleanup immediately after a successful HTTP response, reads the upstream error body before status handling, and returns the actual upstream response text instead of formatting the body object pointer.

## Type of change

# How Has This Been Tested?

- [x] `cd llm/llm-server && go test ./services_server -run TestServiceServerErrorResponsesCloseBody -count=1`

# Maintainer Check Notes

The focused test replaces `common.HttpClient().Transport` with a mocked round tripper and exercises all six early-return cases:

- `ExecuteQuery`: 401 and 500
- `ExecuteScanImageQuery`: 401 and 500
- `ExecuteScanCisQuery`: 401 and 500

For each case it verifies:

- the function returns an error
- the error includes the upstream response body text
- the response body `Close()` method was called

Full package tests include existing integration-style tests that require live service/env configuration, so the narrow unit test above is the fastest deterministic check for this leak fix.

### DIFF
--- a/llm/llm-server/services_server/service.go
+++ b/llm/llm-server/services_server/service.go
@@ -31,22 +31,19 @@ func ExecuteQuery(serviceRequest ServicesQueryRequest) (map[string]string, error
 		return response, fmt.Errorf("services: executequery, unable to process request: %v", err)
 	}
 
-	if resp.StatusCode == 401 {
-		return response, fmt.Errorf("unauthorized: %v", resp.Body)
-	}
+	defer closeResponseBody(resp.Body, "services_server")
 
-	if resp.StatusCode == 500 {
-		return response, fmt.Errorf("internal Server Error from Services Server, %v", resp.Body)
-	}
-
-	defer func() {
-		if err := resp.Body.Close(); err != nil {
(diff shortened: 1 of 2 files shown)


shivam2003-dev:
Closing in favor of #445 with a clean branch name.

PR #444, opened 2026-06-26
ci(benchmark): run lint and tests

### LINKED ISSUE #429: ci(benchmark): Benchmark-Server CI installs deps but never lints or tests
### Gap

`.github/workflows/benchmark-server.yaml` sets up Python + uv and installs dependencies, but has **no lint or test step**. So a PR touching `llm/benchmark/**` gets a green "Benchmark-Server CI" check that only proves *dependencies install* — not that the code is formatted, lint-clean, or that its tests pass. Meanwhile the service is fully tooled for it: `black`/`flake8`/`pytest` are pinned in `pyproject.toml`, there's a `.flake8`, and real tests exist under `benchmark_server/tests/` (auth/JWT/magic-link).

### Why it's not a one-line add

The code is not currently green against its own configured tools (measured with the pinned versions):
- `black --check .` → **28 files** would be reformatted (includes service code, e.g. `benchmark_server/controllers/agent_benchmark_controller.py`, `benchmark_server/utils/run_manager.py`).
- `flake8 .` → **288 findings**. The bulk are in `llm/agents/nubi/fixtures/*/generate_logs.py` (throwaway log-generator scripts: E501/B007/E231/B950), but some are in real code (F401 unused import, F541 empty f-string).
- `pytest` → `benchmark_server/tests/` are standalone unit tests, but the wider `llm/agents/**` tests (e.g. ragas evaluation) pull heavy deps and likely need secrets/network, so the suite needs scoping for CI.

So enabling a gate requires a cleanup + a couple of scoping decisions, not just adding steps.

### Proposed approach (incremental, smallest first)

1. **Lint, scoped to service code** — `black --check` + `flake8` over `benchmark_server/` (and exclude the `llm/agents/nubi/fixtures` log-generators in `.flake8`, joining the existing excludes). Reformat/fix that subset in the same PR so the gate starts green.
2. **Unit tests** — run `pytest benchmark_server/tests/` only (the dependency-light, secret-free suite); leave the `llm/agents` evaluation tests out of PR CI.
3. (Optional, later) widen lint to the full tree once the fixtures are cleaned or excluded.

Happy to send the PR for step 1+2 once the scope (which dirs to lint, which tests to gate) is agreed.

shivam2003-dev:
# Description

Fixes #429

Benchmark-Server CI installed Python dependencies but did not run any benchmark server quality gate. This PR makes the workflow fail on formatting, lint, and unit-test regressions for the dependency-light benchmark server path.

Changes:

- adds `python -m black --check benchmark_server`
- adds `python -m flake8 benchmark_server`
- adds `python -m pytest benchmark_server/tests`
- formats the benchmark server subset with Black so the new gate starts green
- fixes small flake8 findings: import ordering, unused locals/imports, and one long error detail
- adds narrow `.flake8` per-file ignores for existing legacy complexity and long HTML-template lines, avoiding risky controller/template rewrites in a CI-enablement PR

## Type of change

# How Has This Been Tested?

- [x] `cd llm/benchmark && .venv/bin/python -m black --check benchmark_server`
- [x] `cd llm/benchmark && .venv/bin/python -m flake8 benchmark_server`
- [x] `cd llm/benchmark && .venv/bin/python -m pytest benchmark_server/tests`

# Maintainer Check Notes

The workflow commands use `python -m ...` after the existing `uv pip install --system -r uv.lock` and `uv pip install --system -e .` install steps. To review quickly:

1. Check `.github/workflows/benchmark-server.yaml` for the three new post-install steps.
2. Check `.flake8` for scoped ignores. They are limited to pre-existing high-complexity legacy functions and long HTML template strings.
3. Spot-check `agent_benchmark_controller.py` and `orchestrator.py` for the non-format lint fixes.
4. Re-run the exact three commands above from `llm/benchmark`.

The broader `llm/agents/**` benchmark/evaluation tests remain out of this PR gate because they can require secrets, network, or provisioned infrastructure.

### DIFF
--- a/.github/workflows/benchmark-server.yaml
+++ b/.github/workflows/benchmark-server.yaml
@@ -49,3 +49,15 @@ jobs:
 
           # Install in editable mode
           uv pip install --system -e .
+
+      - name: Check formatting
+        working-directory: ./llm/benchmark
+        run: python -m black --check benchmark_server
+
+      - name: Run lint
+        working-directory: ./llm/benchmark
+        run: python -m flake8 benchmark_server
+
+      - name: Run unit tests
+        working-directory: ./llm/benchmark
+        run: python -m pytest benchmark_server/tests
--- a/llm/benchmark/.flake8
+++ b/llm/benchmark/.flake8
@@ -4,3 +4,9 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9
 exclude = .git,__pycache__,docs/source/conf.py,old,build,dist,.venv,.ml-env,notebooks,data
 extend-ignore = E203, W503
+per-file-ignores =
+    benchmark_server/controllers/agent_analyzer_controller.py:C901
+    benchmark_server/controllers/agent_benchmark_controller.py:C901
+    benchmark_server/orchestrator.py:C901
+    benchmark_server/utils/email_templates.py:E501
+    benchmark_server/utils/llm_client.py:C901
--- a/llm/benchmark/benchmark_server/controllers/agent_benchmark_controller.py
+++ b/llm/benchmark/benchmark_server/controllers/agent_benchmark_controller.py
@@ -1,29 +1,30 @@
-import asyncio
-import json
-import logging
 import os
-import sys
-import re
-import signal
-import time
-from datetime import datetime, timedelta, timezone
-from pathlib import Path
-from typing import List, Optional, Union
 
-from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException
-from pydantic import BaseModel
+os.environ["RAGAS_DO_NOT_TRACK"] = "true"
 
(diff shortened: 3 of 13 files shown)


shivam2003-dev:
Addressed the Gemini telemetry feedback. `RAGAS_DO_NOT_TRACK` is now set immediately after importing `os`, before any imports that may load Ragas, and the later duplicate assignment was removed via that move.

Checks run:
- `.venv/bin/python -m black --check benchmark_server`
- `.venv/bin/python -m flake8 benchmark_server`
- `.venv/bin/python -m pytest benchmark_server/tests`
- `python3 -m compileall benchmark_server/controllers/agent_benchmark_controller.py`
- `git diff --check`

blue4209211 (approved):

PR #445, opened 2026-06-26
fix(llm): close services response bodies on error paths

### LINKED ISSUE #434: 🐛 [BUG] -  HTTP Response-Body / Connection Leak on Error Paths in services_server

shivam2003-dev:
# Description

Fixes #434.

`services_server.ExecuteQuery`, `ExecuteScanImageQuery`, and `ExecuteScanCisQuery` could return before closing `resp.Body` on upstream error statuses. During repeated upstream auth failures or server failures, those paths could leak response bodies and keep HTTP connections/file descriptors open.

This PR centralizes response-body reading/closing immediately after a successful HTTP response and applies consistent non-2xx handling:

- 2xx responses continue through normal JSON parsing.
- 401 and other 4xx responses return the upstream detail where useful.
- 5xx responses return a generic status error to avoid exposing upstream internals.
- Any non-2xx status is handled before JSON unmarshalling, so unexpected status codes do not fall through into response parsing.

## Type of change

# How Has This Been Tested?

- [x] `cd llm/llm-server && go test ./services_server -run 'TestReadAndCloseServicesResponseBody|TestServicesHTTPStatusError'`
- [x] `cd llm/llm-server && git diff --check`

# Maintainer Check Notes

The focused tests verify that response bodies are always closed after reading and that status handling differentiates client-side and server-side errors. The previous test mutated the shared HTTP transport and service endpoint; the new tests cover the response helpers directly to avoid shared global state in the test suite.

`go test ./services_server` still fails locally in unrelated existing tests that require relay/service test infrastructure (`TestServiceDependencyGraph`, `TestServiceQueryLogs`). The focused tests above are deterministic for this PR scope.

### DIFF
--- a/llm/llm-server/services_server/service.go
+++ b/llm/llm-server/services_server/service.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"maps"
+	"net/http"
 	"nudgebee/llm/common"
 	"nudgebee/llm/config"
 	"nudgebee/llm/relay"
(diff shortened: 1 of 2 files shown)


shivam2003-dev:
Addressed the Gemini feedback on upstream HTTP error handling and tests.

Changes made:
- Added shared helpers to read/close response bodies and validate HTTP status codes before unmarshalling.
- 5xx errors now return a generic status error instead of exposing upstream response body text.
- Other non-2xx statuses are handled explicitly so they cannot fall through into JSON parsing.
- Reworked the body-close tests to avoid mutating the shared HTTP transport and service endpoint.

Checks run:
- `go test ./services_server -run 'TestReadAndCloseServicesResponseBody|TestServicesHTTPStatusError'`
- `git diff --check`

Note: `go test ./services_server` still fails locally in unrelated existing relay-dependent tests (`TestServiceDependencyGraph`, `TestServiceQueryLogs`), so I kept validation focused on the changed behavior.

blue4209211 (approved):

PR #456, opened 2026-06-26
Exclude event analysis spend from chat budget usage

### LINKED ISSUE #290: [BUG] - AI chat budget runs out early because event-analysis spend is charged to it

shivam2003-dev:
Fixes #290.

## What changed
- Updated the `user_investigation` budget SQL filter to exclude `event-` sessions.
- Preserved the existing `investigation` filter for event-analysis sessions.
- Added a regression test that verifies cost and conversation-count queries use disjoint event and user-investigation filters.

## Root cause
The budget enforcement path routed event sessions and chat sessions as separate modules, but the usage filter for `user_investigation` was empty. That made chat budget usage include every conversation, including event-analysis sessions already counted under `investigation`.

## Maintainer verification
1. `cd llm/llm-server`
2. `go test ./budget`
3. `go test ./budget ./api`
4. With `golangci-lint` installed, run `make lint`.

## Local checks
- PASS `go test ./budget`
- PASS `go test ./budget ./api`
- PASS `git diff --check`

Note: `make lint` could not run locally because `golangci-lint` is not installed in this environment.

### DIFF
--- a/llm/llm-server/budget/service_test.go
+++ b/llm/llm-server/budget/service_test.go
@@ -4,9 +4,14 @@ import (
 	"fmt"
 	"testing"
 
+	"nudgebee/llm/common"
 	"nudgebee/llm/config"
+	"nudgebee/llm/events"
 
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/jmoiron/sqlx"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // saveAndRestoreConfig saves the current config and returns a cleanup function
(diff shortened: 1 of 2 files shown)


shivam2003-dev:
Addressed the Gemini feedback on the budget test. The SQL mock expectations now derive the `event-` pattern from `events.SessionIdPrefixEvent` with regex escaping instead of hardcoding the literal prefix.

Checks run:
- `go test ./budget`
- `go test ./budget ./api`
- `git diff --check`

hsundar-git:
Duplicate #370

Issue #282, opened 2026-06-21
feat(workflow): Call Workflow action should allow selecting a workflow version, not always Live


blue4209211:
## Summary

The **Call Workflow** action always invokes the **Live** version of the target workflow. This is problematic when workflow versioning is in use:

- The **Published / Stable** version contains production-ready logic.
- The **Live** version is being actively edited or tested.
- A caller workflow that uses **Call Workflow** silently picks up the in-flight Live version — which may contain untested changes — and runs them in production.

The caller should be able to pin the version it invokes (e.g. "always use Published", or pin a specific version) so that production-bound callers don't break when someone edits the callee's Live draft.

## Expected behaviour

The Call Workflow action editor should expose a **version selector** alongside the workflow picker:

- **Live** (current default — keep working for backwards compatibility)
- **Published / Stable** (recommended for production callers)
- **Pinned to version N** (explicit pin to a specific historical version)

At execute time, the runbook engine resolves the selected version to the matching workflow definition and runs it.

## Reproduction

1. Create workflow `W1` and publish a Stable version.
2. Edit `W1` further without publishing — Live now diverges from Stable.
3. Create workflow `W2` with a **Call Workflow** action targeting `W1`.
4. Run `W2`.
5. **Observed:** `W2` executes `W1`'s Live (in-progress) version.
6. **Expected:** `W2` executes whichever version `W2`'s Call Workflow action specified — default Live for backwards compat, but selectable.

## Where to start (rough orientation for a contributor)

- **Frontend (`app/`):** the Call Workflow action editor lives in the workflow-builder UI. Look for the `Call Workflow` task type's configuration component — it currently exposes a workflow picker; needs a sibling version picker.
- **Backend (`runbook-server/`):** the executor that resolves the Call Workflow task needs to honor a `version` field in the task config (default to Live when absent). Look for the task type's executor in `runbook-server/internal/tasks/`.
- **Persistence:** the workflow_versions table already exists; the resolver just needs to read the version selector from the task config instead of hard-coding "Live".

(Verify the exact paths in your checkout — these are rough pointers.)

## Acceptance

After the change:
1. Editing a Call Workflow action shows a version selector. Existing actions default to Live (no migration needed).
2. Running a workflow with a Call Workflow action pinned to Stable executes the published Stable version, even if Live has uncommitted changes.
3. Pinning to a specific historical version works the same way.

jenistenxavier:
Picking this up. One design question before I build:

[...]

Issue #258, opened 2026-06-19
feat(integrations): sync Grafana alerts into events (reconcile, not just webhook)


mayankpande88:
## Summary

Keep NudgeBee's event view of Grafana alerts consistent with Grafana, even when webhooks are missed. Today Grafana alerts only arrive via the inbound webhook (push), which is lossy — a missed firing/resolved leaves events diverged from Grafana's real state. Add a periodic **sync** that pulls Grafana's current alerts and reconciles them into the event-rules pipeline.

## Depends on

- #259 — registers the agent-reported Grafana connection as an integration (provides the connection/auth record this sync uses)
- #257 — the shared `mapGrafanaAlertToEvent` mapping, so the sync and the webhook produce identical, fingerprint-keyed events.

## What to do

1. **Connection/auth**: use the Grafana integration record from #259 (the agent-reported connection registered as an integration) for the base URL + auth. (Until #259 lands there is no integration to query against.)
2. **Query**: pull current firing alerts via the existing Grafana proxy (e.g. `GET /api/prometheus/grafana/api/v1/alerts`). No agent change — the proxy forwards arbitrary GETs.
3. **Reconcile** (scheduled): for each firing alert, map via the shared function (#257) and upsert the event, deduped by `fingerprint`; for open Grafana-sourced events whose alert is no longer firing, resolve them.
4. Constrain to **read-only GET** alerting endpoints (the proxy would forward anything — don't expose mutating Grafana APIs).

## Done when

- [ ] A scheduled sync pulls Grafana's current alerts and reconciles them into events by fingerprint
- [ ] Missed firings show up as events; cleared alerts get resolved
- [ ] No duplicates versus webhook-delivered events (shared mapping + fingerprint)
- [ ] Read-only; clear errors when Grafana is unreachable or unauthorized

PR #458, opened 2026-06-26
feat(llm-server): surface executed command in sub-agent observations

### LINKED ISSUE #292: Sub-agent observations should surface the executed command, not just the natural-language tool input
## Problem

The final-answer LLM for any agent that delegates to a sub-agent only sees:

- Its own natural-language `tool_input` (e.g. `"list pods in the demo namespace"`)
- The sub-agent's formatted output

It does **not** see the actual shell command the sub-agent executed (e.g. `kubectl get pods -n <ns>`). This blocks every Soul-style preference that asks the model to "lead with the command" (`format.codeFirst`), and makes troubleshooting / audit harder — you can read the output but not the exact command that produced it.

Verified live: the final-answer prompt contained `<user_style>format_code_first: true</user_style>` *and* the pod-list table, but no shell command anywhere. The `k8s_debug` top-level model could not surface the command because it never received it.

## Generic to all agents

Same shape on every delegated tool path — kubectl, aws, gcp, azure, loganalysis, anything that runs a shell or API call via a sub-agent. Whatever the sub-agent actually executed is opaque to its caller.

## Proposed fix

Carry the executed command(s) back from the sub-agent and stamp them on the `<observation>` element the caller sees. Something like:

```xml
<observation step="E1" tool="kubectl" command="kubectl get pods -n <ns>">
| Pod Name | ...
</observation>
```

For multi-step sub-agents, surface the list of commands run.

Implementation lives near where `llm/llm-server/agents/core/executor_planner.go` writes the observation block to the caller's scratchpad. Sub-agent results already carry the tool invocations internally; they just aren't propagated up.

## Acceptance

- For at least kubectl, aws, gcp, azure delegated calls, the caller's prompt contains the actual executed command alongside the output.
- A Soul-driven `format.codeFirst` chat (e.g. "list pods in &lt;ns&gt;") leads with the `kubectl …` command in the final answer.

jenistenxavier:
# Description

Sub-agent tool observations rendered stdout only. The actual command dispatched to the workspace pod — after the tool's input transforms (`aws`/`gcloud` prefix injection, `--context` strip, `.nb_profile` sourcing) — was discarded. The planner could not see or repeat exactly what ran, weakening grounding and code-first answer formatting.

This threads the executed command through the existing `NBToolResponseMetadata` seam so it appears alongside each observation.

Part of #292

> **Scope:** this PR lands the **leaf-level foundation** — leaf tools stamp `ExecutedCommand` and the centralized footer renders it. It does **not** yet satisfy acceptance #1 for the delegated (ReWOO → ReAct sub-agent) path: `nbAgentTool.Call` drops `Metadata` on every return branch, so the orchestrator's prompt still gets no command for delegated steps. That propagation is tracked as a follow-up checklist on #292, which stays open after this merge. (`Part of`, not `Fixes`, is intentional — do not let merge auto-close #292.)

## What changed

- **`tools/core/interface.go`** — new `NBToolResponseMetadata.ExecutedCommand` field. Persists automatically via the existing Metadata JSONB column (V751); no migration.
- **`executor_planner.go`** — `formatToolMetadataFooter` now renders `[exitStatus: N | executionDuration: Xms | command: <cmd>]`. Single centralized seam → all render paths (scratchpad ×2, react_2, react_3) inherit it. Command capped at 200 chars via `TruncateMiddle` so long jq/pipe chains don't bloat compressed older turns (full command still persisted).
- **Tools** — `shell`, `kubectl`, `aws`, `gcp`, `azure` stamp `ExecutedCommand` on success and no-match return paths. `helm` updated for the `successResponseNoMatches` signature change.

## Security

The **clean, pre-auth-wrap** command is stored (never the `WrapCommandWithAuth` output), and run through `ScrubCredentials` before being set — mirroring the existing output-scrub. Error paths already surface the command via `wrapShellError`/`wrapKubectlError`, so they are intentionally not double-stamped.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] `go build ./...` + `go vet` clean
- [x] `gofmt` clean
- [x] New unit tests: footer command rendering, 200-char truncation, `successResponseNoMatches` metadata (incl. credential-scrub assertion) — all pass
- [x] Existing footer/render tests pass unchanged (no-command output is byte-identical)

## Review Notes — Risks & Counterarguments

- **Partial redundancy with `<action><tool_input>`.** For shell/aws/gcp the executed command is close to the LLM's raw input. Value lives in the transform diff (prefix injection, context strip) and co-locating the command *in the observation*. A dedup-vs-ToolInput guard was considered and dropped: `ToolInput` is rewritten by turn-cache normalization and has an inconsistent JSON/raw shape, making the comparison fragile.
- **Scope is incremental.** Only the 5 issue-named CLI tools are stamped this PR, matching the established per-tool rollout precedent (`Metadata.Stderr` shipped kubectl-only). Remaining `ExecuteOrLazyCreate` tools (clickhouse, mysql, redis, argocd, github, …) follow the same one-line pattern in a follow-up.
- **Token cost.** Footer rides every observation incl. compressed turns; mitigated by the 200-char cap.


runbook-server/internal/tasks/core/call_workflow_task.go
```diff
@@ -15,6 +16,54 @@ import (
 	"go.temporal.io/sdk/client"
 )
 
+// resolveTargetVersion picks which snapshot of the callee to run. When the task
+// config carries a positive integer `workflow_version`, it pins to that specific
+// historical version (GetWorkflowVersion); otherwise it follows the callee's
+// floating Live pointer (GetLiveWorkflowVersion). The Live fallback is the
+// backwards-compatible default for every existing Call Workflow action — absent
+// param == today's behavior, so no migration is needed (#282).
+func resolveTargetVersion(ctx context.Context, store model.WorkflowStore, workflowID, workflowName string, params map[string]any) (*model.WorkflowVersion, error) {
```
hsundar-git, line 26: medium

This file (plus `CallWorkflowFields.tsx` and `call_workflow_task_test.go`) implements **workflow version pinning** (`#282`/`#258`) — a feature unrelated to this PR's stated scope, *surface executed command in sub-agent observations* (`Fixes #292`). The PR title (`feat(llm-server)`), description, and linked issue cover only the llm-server change; this runbook-server + frontend feature is undocumented here.

Bundling two independent features across three services makes the diff harder to review, couples their risk, and muddies traceability (a reviewer of #292 wouldn't expect version-pinning code). The version-pinning code itself looks well-built and tested — the concern is purely that it should be its **own PR** linked to #282.

Recommend splitting into two PRs.

llm/llm-server/tools/core/interface.go
```diff
@@ -98,6 +98,13 @@ type NBToolResponseMetadata struct {
 	// callers can tell the planner how much was dropped.
 	Truncated   bool `json:"truncated,omitempty"`
 	OriginalLen int  `json:"original_len,omitempty"`
+	// ExecutedCommand is the actual command dispatched to the workspace pod,
+	// after the tool's input transforms (e.g. "aws " prefix, --context strip,
+	// profile sourcing). It surfaces the real command behind a sub-agent's
+	// observation so the planner can ground its answer on what was run.
+	// MUST be credential-scrubbed by the tool before it is set (the auth-wrap
+	// form is never stored). Empty for tools that don't shell out.
+	ExecutedCommand string `json:"executed_command,omitempty"`
```
hsundar-git, line 111: low

The doc says this field **"MUST be credential-scrubbed by the tool before it is set"**, but the invariant is only enforced on some stamp sites: `shell` and the AWS/Azure/GCP **workspace** paths call `ScrubCredentials(...)`, while `kubectl`, `helm`, and the AWS/Azure/GCP **non-workspace** (`cloud.Execute`) paths set the raw `command`.

In practice those raw paths are currently safe — the `cloud.Execute` path injects no creds into the command, and kubectl/helm commands don't carry secrets — so this isn't a live leak. But the gap means a future tool/path that *does* embed a secret would silently violate the documented contract.

Suggest either (a) scrub uniformly at every stamp site, or (b) soften the comment to "scrubbed wherever the tool injects credentials into the command" so the doc matches reality.

hsundar-git:
## Summary of Changes

This PR's stated goal (`Fixes #292`) is to **surface the executed command** in sub-agent tool observations: a new `NBToolResponseMetadata.ExecutedCommand` field is threaded through the existing metadata seam, rendered into the observation footer as `[exitStatus | executionDuration | command]`, and stamped by the shell/kubectl/aws/gcp/azure/helm tools (credential-scrubbed on the credential-injecting paths).

It **also** contains a second, independent feature — **Call Workflow version pinning** (`#282`/`#258`) — adding a `workflow_version` resolver in `runbook-server` and a version-selector dropdown in the frontend, which the title/description do not mention.

### Highlights
- **Executed-command footer** — single seam in `formatToolMetadataFooter`; 200-char `TruncateMiddle` cap; full command persisted in the Metadata JSONB. Clean, well-tested.
- **kubectl metadata nil-safety** — `Metadata` is now always constructed, and `Stderr` is added to it instead of overwriting. Good incidental fix.
- **(Unrelated) workflow version pinning** — `resolveTargetVersion` + `parseWorkflowVersionParam` with solid coercion/guard rails and backwards-compat (absent → Live). Genuinely good code — just out of scope for this PR.

### Review Summary

**Services affected:** llm-server (Go), runbook-server (Go), frontend (TypeScript)
**Risk level:** Low (code is sound; main concern is PR hygiene)

|---|---|---|
| Scope / process | `call_workflow_task.go` + `CallWorkflowFields.tsx` ship an unrelated feature (#282) not described in the PR; recommend splitting — `call_workflow_task.go:25` | medium |
| Security / consistency | `ExecutedCommand` doc says "MUST be credential-scrubbed", but kubectl/helm + the aws/azure/gcp `cloud.Execute` paths stamp the raw command (safe today, but breaks the documented invariant) — `tools/core/interface.go:107` | low |
| Correctness | Footer truncation, command capping, and nil-safe metadata all look correct | low |

> ✅ **Looks Good** — Test coverage is strong on both features (footer rendering, 200-char truncation, no-match metadata + scrub assertion; version pin/default-to-live/param coercion). Credential handling on the workspace paths correctly stores the clean pre-auth-wrap command.

### Checklist
- [x] Tests cover new/changed behavior
- [x] No hardcoded secrets
- [x] Error handling is appropriate
- [x] No breaking API changes (new param is optional; absent → prior behavior)
- [ ] **PR scope** — two unrelated features in one PR; split recommended

---
*Automated PR Review*

runbook-server/internal/tasks/core/call_workflow_task.go
jenistenxavier: Done — split out. Reverted the version-pinning commit (2e167916) so this PR now contains only the #292 `surface executed command` change. The version-pin feature lives in its own PR (#457, linked to #282).

llm/llm-server/tools/core/interface.go
jenistenxavier: Fixed in 49d00bff (option b). Reworded the doc to match reality: scrubbing is required only on paths that inject credentials into the command string (shell + AWS/Azure/GCP workspace-exec). kubectl/helm and the cloud.Execute backend paths carry no inline secrets (auth applied server-side), so they store the command as-is.

jenistenxavier:
@gemini-code-assist Question on scope before we finalize this PR.

**Context:** This PR stamps `NBToolResponseMetadata.ExecutedCommand` on the leaf tools (`shell`, `kubectl`, `aws`, `gcp`, `azure`) and renders it via the centralized `formatToolMetadataFooter` footer. That fully works for **top-level ReAct agents**, where the executor runs the leaf tool directly → footer lands in the scratchpad the planner/solver sees.

**The gap I want your read on:** Issue #292 is specifically about the **delegate-to-sub-agent** path, and its acceptance #1 says *"for at least kubectl, aws, gcp, azure delegated calls, the **caller's prompt** contains the actual executed command."*

For a ReWOO orchestrator like `k8s_debug` (`agent_k8s_debug_2.go:140`), the tools are sub-agents (`KubectlAgent`, etc. — `agent_k8s_debug_2.go:222`), and `KubectlAgent` is itself ReAct (`agent_kubectl.go:187`) which calls the leaf `tool_kubectl`. So the flow is:

```
k8s_debug (ReWOO, the CALLER)
  └─ kubectl sub-agent (ReAct, via nbAgentTool.Call)
       └─ tool_kubectl (leaf)  ← ExecutedCommand stamped here
```

The leaf-stamped command renders in the **sub-agent's own** scratchpad, but `nbAgentTool.Call` returns to the ReWOO caller with **`Metadata` unset on all return branches** (`factory_agent.go:188,201,223,234,247,259`). So the delegated step's `step.Metadata` is nil at the ReWOO level and the caller's prompt gets no command footer. The command only reaches the final answer if the sub-agent happens to echo it in its answer text (prompt-dependent, not structural).

**Questions:**
1. Do you agree this PR, as-is, does **not** satisfy acceptance #1 for the delegated (ReWOO → ReAct sub-agent) path that the issue's live repro describes?
2. Is wiring the propagation through `nbAgentTool.Call` (pull executed commands from `resp.AgentStepResponse`, set `Metadata.ExecutedCommand` on the returned `NBToolResponse` — single command, or joined list for multi-step) in-scope for **this** PR, or is it cleaner as a follow-up with #292 kept open and this PR relabeled as the leaf-stamping + render-seam foundation?

Looking for a scope call, not a rewrite. Thanks.

llm/llm-server/agents/core/executor_planner.go
```diff
@@ -1535,7 +1535,20 @@ func formatToolMetadataFooter(metadata *toolcore.NBToolResponseMetadata) string
...
+	footer := fmt.Sprintf("\n[exitStatus: %d | executionDuration: %dms", metadata.ExitStatus, ms)
+	if cmd := metadata.ExecutedCommand; cmd != "" {
+		// Cap the rendered command so a long jq/pipe chain can't bloat every
+		// observation (incl. compressed older turns). Keep both ends: the head
+		// names the binary, the tail carries namespace/region/filters. The full
+		// command is still persisted in the Metadata JSONB column.
+		const maxCmdChars = 200
+		if len(cmd) > maxCmdChars {
+			cmd = TruncateMiddle(cmd, 150, 50)
+		}
```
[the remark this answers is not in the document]
jenistenxavier: Fixed in 618a185d. Switched the cap to `utf8.RuneCountInString` (renamed const to `maxCmdRunes`). `TruncateMiddle`/`TruncateHead`/`truncateTail` were already rune-boundary safe (they walk `utf8.RuneStart`), so there was no malformed-UTF-8 risk — but the threshold check was on byte length, which could truncate multi-byte commands early. Added `TestFormatToolMetadataFooter_CommandRuneCounting`: a 150-rune (3-byte 'é') command stays under the cap untouched, a 300-rune 4-byte emoji command is truncated, both asserted `utf8.ValidString`.

jenistenxavier:
Agreed — keeping this PR scoped to the leaf-level foundation. Changed the PR body from `Fixes #292` to `Part of #292` so the merge won't auto-close the issue, and dropped your three guidance points (`nbAgentTool.Call` extraction, multi-step aggregation, scrub-invariant check) as a follow-up checklist on #292. The delegated ReWOO → ReAct propagation lands in a separate PR.

PR #473, opened 2026-06-27
fix(ui): reset Create Ticket loading state on validation failure

### LINKED ISSUE #472: fix(ui): Create Ticket button stuck in loading state after required-field validation fails
## Describe the bug

In the ticket creation modal (`app/src/components1/tickets/TicketCreatePopupForm.jsx`), the **Create Ticket** button can get permanently stuck in its loading/disabled state.

`createTicket()` sets `setIsLoading(true)` *before* running the required-custom-fields validation. When that validation fails, the handler shows an error toast and `return`s early — but it never resets `setIsLoading(false)`. The only place loading is cleared is the `.finally()` of the API call, which is never reached on this early-return path.

`setIsLoading(false)` is therefore only run on:
- the API success/error path (`.finally()`), or
- closing/cancelling the modal (`handleCancel`).

So after a required-field validation error, the user is left with a spinning, disabled **Create Ticket** button and cannot retry without closing the modal.

## Steps to reproduce

1. Open the Create Ticket modal for an integration whose issue type has a **required custom field** (e.g. a Jira project with a mandatory field).
2. Fill the basic fields but leave a required custom field empty.
3. Click **Create Ticket**.
4. A toast appears ("<field> cannot be empty") — but the button stays in the loading state and is disabled.

## Expected behavior

When validation fails, the button should return to its normal (non-loading) state so the user can correct the field and retry.

## Component / Area

- UI / Tickets

bharathnr1:
# Description

In the ticket creation modal (`app/src/components1/tickets/TicketCreatePopupForm.jsx`), `createTicket()` called `setIsLoading(true)` **before** the required-custom-fields validation. When that validation failed, it showed an error toast and returned early — without resetting `setIsLoading(false)`. Loading is only cleared in the API call's `.finally()` (never reached on that path) or on modal cancel, so the **Create Ticket** button was left permanently spinning and disabled, forcing the user to close the modal to retry.

This moves the single `setIsLoading(true)` to **after** validation passes (immediately before the API call). The code between the old and new positions is purely synchronous (a `JSON.parse`/`stringify` clone and date formatting), so nothing relied on the early flag.

Fixes #472

## Type of change

# How Has This Been Tested?

- [x] Prettier (`npx prettier@2.8.8 --check`) clean on the changed file
- [x] Manual trace of the handler: validation-failure path now returns with `isLoading=false`; success/error paths still clear it via `.finally()`; cancel still clears it via `handleCancel`.

- The removed `setIsLoading(true)` was redundant with the existing one right before the API call; there was already a duplicate set-true. Net behavior change is only on the validation-failure early-return path.
- No new state, no API/contract change — single-line move, smallest possible fix.

### DIFF
--- a/app/src/components1/tickets/TicketCreatePopupForm.jsx
+++ b/app/src/components1/tickets/TicketCreatePopupForm.jsx
@@ -33,7 +33,6 @@ const AddModalForm = ({
     const { selectedConfig, selectedProject, selectedIssueType, ticketDetails, formData, selectedIssueTypeTicketMetadata } = ticketState;
 
     setError(false);
-    setIsLoading(true);
     let cloneObj = JSON.parse(JSON.stringify(formData));
     delete cloneObj.assignee;
     for (const [_key, value] of Object.entries(selectedIssueTypeTicketMetadata?.[0]?.fields ?? [])) {


saiprasadpotdar (commented):
LGTM!

saiprasadpotdar (approved):
LGTM

Issue #447, opened 2026-06-26
Security: service-token auth should fail closed when token env vars are unset


shivam2003-dev:
## Security score
9/10

Service-token authentication can fail open when shared token env vars are missing, which can expose internal service-to-service routes unexpectedly.

## References
- [CWE-636: Not Failing Securely (Failing Open)](https://cwe.mitre.org/data/definitions/636.html)
- [OWASP: Fail Securely](https://owasp.org/www-community/Fail_securely)
- [OWASP Authorization Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authorization_Cheat_Sheet.html)
- [CWE-306: Missing Authentication for Critical Function](https://cwe.mitre.org/data/definitions/306.html)

## Summary

Several service-to-service auth gates disable themselves or match an empty shared token when the expected token env var is unset. That makes internal RPC/admin surfaces depend entirely on network isolation in misconfigured or custom deployments.

## Evidence

- `llm/llm-server/cmd/main.go`: `authHandlerMiddleware` only checks `X-ACTION-TOKEN` when `config.Config.LlmServerToken != ""`; if `LLM_SERVER_TOKEN` is empty, protected routes continue without a token.
- `llm/llm-server/cmd/auth_test.go`: the test case `token unset: any header value passes` pins this behavior for `/api/admin/prompts/config` and `/v1/llm-config/test-connection`.
- `deploy/kubernetes/nudgebee/templates/secrets-nudgebee.yaml`: `LLM_SERVER_TOKEN` defaults to `""`, and the comment documents that the llm-server gate is lenient for in-cluster traffic.
- `api-server/services/cmd/main.go`: `authHeader == config.Config.ServiceApiServerToken` means an empty `ACTION_API_SERVER_TOKEN` accepts requests without the header; the startup warning explicitly says every endpoint is effectively unauthenticated.
- `collector-server/cloud-collector/cmd/main.go`: same direct equality check exists for `CLOUD_COLLECTOR_SERVER_TOKEN`, so an empty expected token also accepts a missing header.

## Impact

If any of these token env vars are omitted, rendered empty by GitOps/offline Helm workflows, or lost during a custom deployment, callers that can reach the service ClusterIP or any exposed ingress can hit internal RPC/admin endpoints without the intended shared-secret proof. For llm-server this includes admin prompt/config surfaces and LLM config test routes; for api-server/cloud-collector this can open broader internal action surfaces.

[...]

PR #481, opened 2026-06-27
security(server): protect pprof debug endpoints across services

### LINKED ISSUE #449: Security: protect debug pprof endpoints across services
## Security score
7/10

Unauthenticated debug pprof endpoints can leak runtime details and support denial-of-service profiling attacks, but exploitation depends on network exposure.

## References
- [Go `net/http/pprof` package documentation](https://pkg.go.dev/net/http/pprof)
- [CWE-489: Active Debug Code](https://cwe.mitre.org/data/definitions/489.html)
- [CWE-200: Exposure of Sensitive Information to an Unauthorized Actor](https://cwe.mitre.org/data/definitions/200.html)
- [NVD CVE-2019-11248: unauthenticated Kubernetes kubelet pprof exposure](https://nvd.nist.gov/vuln/detail/CVE-2019-11248)
- [Bearer rule: Usage of active debug code, pprof enabled](https://docs.bearer.com/reference/rules/go_gosec_leak_pprof_endpoint/)

## Summary

Several Go services register `gin-contrib/pprof` debug endpoints without effective authentication. `/debug/pprof/*` can expose heap/goroutine/profile data and can be used for expensive profiling requests. These endpoints should be treated as internal privileged diagnostics, not public health/debug routes.

## Evidence

- `llm/llm-server/cmd/main.go`
  - `pprof.Register(r)` runs before `r.Use(authHandlerMiddleware())`.
  - `authHandlerMiddleware` also explicitly marks any path with prefix `/debug` public.
- `collector-server/cloud-collector/cmd/main.go`
  - `pprof.Register(r)` runs before `r.Use(authHandlerMiddleware())`.
  - `authHandlerMiddleware` explicitly marks any path with prefix `/debug` public.
- `api-server/services/cmd/main.go`
  - `pprof.Register(r)` runs before `r.Use(authHandlerMiddleware())`. Gin snapshots middleware at route registration time, so the later auth middleware does not protect the already-registered pprof handlers.
- `ticket-server/main.go`
  - `pprof.Register(router)` is registered on the root router before normal routes, with no global service-token gate around those debug routes.
- `collector-server/k8s-collector/relay-server/pkg/server/router.go` shows the safer local pattern: `/debug/pprof` is registered on a group protected by `ClientAuthMiddleware`, with comments noting pprof can leak in-flight secrets/tokens.

## Impact

If any affected service is reachable beyond a strictly trusted in-cluster boundary, an unauthenticated caller can access `/debug/pprof/`, `/debug/pprof/goroutine`, `/debug/pprof/heap`, `/debug/pprof/profile`, etc. Heap and goroutine dumps may include sensitive request state, headers, tokens, URLs, integration metadata, or tenant/account identifiers. CPU/profile endpoints can also impose unnecessary load.

This is defense-in-depth relevant even for ClusterIP-only services because compromised pods, accidental ingress exposure, port-forward workflows, or service mesh routing mistakes should not turn diagnostics into unauthenticated introspection.

## Recommended fix

1. Register pprof routes only behind the same service-token middleware used for internal RPC/admin routes, or introduce a dedicated diagnostics token.
2. Do not whitelist all `/debug` paths in auth middleware. Keep `/health` public; require auth for `/debug/pprof/*`.
3. Register pprof after global middleware or on an explicitly protected Gin group, matching the relay-server pattern:
   - `group := r.Group("/debug/pprof", authMiddleware)`
   - `pprof.RouteRegister(group, "")`
4. Add tests proving `/debug/pprof/heap` and `/debug/pprof/goroutine` return 401 without the token and 200 with it.
5. Consider a Helm value to disable pprof entirely by default in production.

bharathnr1:
# Description

`gin-contrib/pprof` registered `/debug/pprof/*` **before** the service-token auth middleware was attached. Gin snapshots the middleware chain at route-registration time, so the later `r.Use(authHandlerMiddleware())` never protected those routes. In `llm-server` and `cloud-collector` the auth middleware *also* explicitly marked any `/debug` path public. Net result: `/debug/pprof/`, `/heap`, `/goroutine`, `/profile`, etc. were served **unauthenticated** — heap/goroutine dumps can leak in-flight request state, tokens, integration metadata, and tenant/account identifiers, and the profile endpoints invite DoS. (CWE-489 / CWE-200.)

## Changes
- **llm-server**, **cloud-collector** (`cmd/main.go`): remove the `/debug` public exemption from `authHandlerMiddleware`, and register `pprof` **after** `r.Use(authHandlerMiddleware())` so it inherits the service-token gate.
- **api-server** (`cmd/main.go`): register `pprof` after the auth middleware. (Its middleware never exempted `/debug`; the bug was purely registration order.)
- **ticket-server** (`main.go`): it has **no** global inbound service-token middleware (it relies on network isolation), so there is no gate to register behind. Gate `pprof` behind `PPROF_ENABLED=true` (default **off**) so debug endpoints are opt-in for diagnostics instead of always-on.

This mirrors the existing correct pattern in `collector-server/k8s-collector/relay-server/pkg/server/router.go`, which already gates pprof behind `ClientAuthMiddleware`.

Verified no non-pprof routes serve under `/debug` in any of these services, so dropping the exemption only affects pprof.

Fixes #449

## Type of change

- [x] Security (non-breaking change which improves security)

# How Has This Been Tested?

- [x] **llm-server** `cmd/auth_test.go` extended: with the token configured, `/debug/pprof/heap` → 401 without the header, 200 with it.
- [x] **cloud-collector** new `cmd/auth_test.go`: `/health` stays public; `/debug/pprof/heap` requires the token (401 missing/wrong, 200 correct).
- [x] `go test ./cmd/` passes for llm-server and cloud-collector; `go vet ./cmd/` clean for api-server; `go vet .` clean for ticket-server; `gofmt` clean.

- **Interplay with #447 (fail-open auth):** for the three Gin services, when the service token env var is unset the middleware is a no-op, so pprof would still be reachable. That fail-open behavior is the subject of #447; this PR implements #449's explicit recommendation (gate pprof behind the *same* service-token middleware). With both merged, pprof is fully closed.
- **ticket-server uses a different mechanism** (env flag, not middleware) because it has no inbound auth layer to reuse; adding a full auth middleware there is out of scope for this issue. Operators who currently rely on pprof must set `PPROF_ENABLED=true`.
- **api-server / ticket-server** changes are router-wiring in `main()` and aren't unit-tested here (would require refactoring `main()` into a testable builder); they're mechanical and `vet`-clean. The middleware-level behavior is covered by the llm-server/cloud-collector tests.

### DIFF
--- a/ticket-server/main.go
+++ b/ticket-server/main.go
@@ -9,6 +9,7 @@ import (
 	"nudgebee/tickets-server/routes"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"github.com/gin-contrib/pprof"
@@ -54,7 +55,15 @@ func main() {
 	}()
 	gin.SetMode(gin.ReleaseMode)
 	router := gin.New()
-	pprof.Register(router)
+	// ticket-server has no global inbound service-token middleware (it relies on
+	// network isolation), so there is no auth gate to register pprof behind.
+	// pprof heap/goroutine dumps can leak in-flight request state, so it must
+	// not be served unauthenticated by default — make it explicit opt-in via
+	// PPROF_ENABLED=true for diagnostics.
+	if strings.EqualFold(os.Getenv("PPROF_ENABLED"), "true") {
+		pprof.Register(router)
+		slog.Warn("pprof debug endpoints enabled at /debug/pprof (PPROF_ENABLED=true) — ensure this service is not externally reachable")
+	}
 	router.Use(gin.Recovery())
 	router.Use(sloggin.NewWithFilters(logger, sloggin.IgnorePath("/health")))
 
--- a/ticket-server/main.go
+++ b/ticket-server/main.go
@@ -9,7 +9,6 @@ import (
 	"nudgebee/tickets-server/routes"
 	"os"
 	"os/signal"
-	"strings"
 	"syscall"
 
 	"github.com/gin-contrib/pprof"
@@ -55,17 +54,14 @@ func main() {
 	}()
 	gin.SetMode(gin.ReleaseMode)
 	router := gin.New()
-	// ticket-server has no global inbound service-token middleware (it relies on
-	// network isolation), so there is no auth gate to register pprof behind.
-	// pprof heap/goroutine dumps can leak in-flight request state, so it must
-	// not be served unauthenticated by default — make it explicit opt-in via
-	// PPROF_ENABLED=true for diagnostics.
-	if strings.EqualFold(os.Getenv("PPROF_ENABLED"), "true") {
-		pprof.Register(router)
-		slog.Warn("pprof debug endpoints enabled at /debug/pprof (PPROF_ENABLED=true) — ensure this service is not externally reachable")
-	}
 	router.Use(gin.Recovery())
 	router.Use(sloggin.NewWithFilters(logger, sloggin.IgnorePath("/health")))
+
+	// ticket-server has no global inbound service-token middleware yet (it relies
+	// on network isolation); the service token will be added later, at which
+	// point pprof inherits the gate like the other services. Registered after
+	// Recovery + logging so the debug routes still get panic recovery and logs.
+	pprof.Register(router)
 
 	routes.InitializeRoutes(router)
 
--- a/api-server/services/cmd/main.go
+++ b/api-server/services/cmd/main.go
@@ -182,12 +182,16 @@ func main() {
 
 	gin.SetMode(gin.ReleaseMode)
 	r := gin.New()
-	pprof.Register(r)
 	r.Use(gin.Recovery())
 	r.Use(sloggin.NewWithFilters(logger, sloggin.IgnorePath("/health")))
 	r.Use(otelgin.Middleware(config.SERVICE_NAME))
 	r.Use(traceResponseHeaderMiddleware())
 	r.Use(authHandlerMiddleware())
+	// Register pprof AFTER the auth middleware so /debug/pprof inherits the
+	// service-token gate. Gin snapshots middleware at registration time, so
+	// registering before r.Use(authHandlerMiddleware()) left the debug
+	// endpoints unauthenticated (heap/goroutine dumps can leak request state).
+	pprof.Register(r)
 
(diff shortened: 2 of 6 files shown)


ticket-server/main.go
```diff
@@ -54,7 +55,15 @@ func main() {
...
+	// network isolation), so there is no auth gate to register pprof behind.
+	// pprof heap/goroutine dumps can leak in-flight request state, so it must
+	// not be served unauthenticated by default — make it explicit opt-in via
+	// PPROF_ENABLED=true for diagnostics.
+	if strings.EqualFold(os.Getenv("PPROF_ENABLED"), "true") {
+		pprof.Register(router)
+		slog.Warn("pprof debug endpoints enabled at /debug/pprof (PPROF_ENABLED=true) — ensure this service is not externally reachable")
+	}
 	router.Use(gin.Recovery())
 	router.Use(sloggin.NewWithFilters(logger, sloggin.IgnorePath("/health")))
```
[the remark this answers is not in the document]
bharathnr1: Good catch — moved the pprof registration to after `gin.Recovery()` and the slog middleware so the debug routes get panic recovery and request logging. Pushed.

ticket-server/main.go
```diff
@@ -54,10 +55,20 @@ func main() {
...
 
+	// ticket-server has no global inbound service-token middleware (it relies on
+	// network isolation), so there is no auth gate to register pprof behind.
+	// pprof heap/goroutine dumps can leak in-flight request state, so it must
+	// not be served unauthenticated by default — make it explicit opt-in via
+	// PPROF_ENABLED=true for diagnostics. Registered after Recovery + logging so
+	// the debug routes still get panic recovery and request logs.
+	if strings.EqualFold(os.Getenv("PPROF_ENABLED"), "true") {
```
blue4209211: we can remove this check as its misisng in other services as well, we will eventually add ticket_Server_Token

blue4209211 (changes requested):

ticket-server/main.go
bharathnr1: Done.

PR #487, opened 2026-06-28
fix(compose): use string form for env_file to support older Docker Compose versions

AruneshDwivedi:
env_file entries used the long-form object syntax (path + required) which requires Docker Compose v2.24+. Switching to the plain string form works across all versions while preserving the documented cp .env.example .env setup step.


AruneshDwivedi:
I have read the CLA Document and I hereby sign the CLA

PR #490, opened 2026-06-28
fix: prevent AI assistant from fabricating metric values on empty query results

### LINKED ISSUE #291: [BUG] - AI assistant sometimes reports made-up metric values when the underlying query returned no data
## Description

When the AI assistant answers a question about a service's metrics (error rate, latency, throughput) and the metric queries it runs return **no data**, it sometimes responds with **specific, made-up numbers** as if they were real — instead of saying the data is unavailable. For example, it reported P95 245ms / P99 412ms / 12.4 req/s for a service while every query it ran came back empty.

This is a correctness/trust problem: an operator could act on performance figures that were never measured.

## Impact

- **Who is affected**: anyone using the AI assistant for metrics dashboards or root-cause investigations.
- **Severity**: fabricated performance numbers presented as fact, with no signal that they're unreliable. Worse than reporting "N/A", because it looks authoritative.
- **Since when**: pre-existing behaviour of the answer-synthesis step; not introduced by a specific recent change. It is **intermittent** — it depends on the model's run-to-run behaviour, so the same question can fabricate on one run and answer honestly on another.

## Reproduction Steps

1. Ask the assistant for a service's RED metrics (error rate / p95 / p99 / throughput) for a service whose metrics are NOT under the naming the assistant tries first (so its initial queries return empty).
2. Observe that the answer can contain specific latency/throughput values that no successful query produced.

Because it's intermittent (model sampling), it may take several runs to reproduce; the honest-vs-fabricated outcome varies per run on the same input.

---

## Technical Details

Observed in a local run for the query *"checkout error rate / p95 / p99 / throughput, last 1h"*:

- The Prometheus agent issued **10 `prometheus_execute` calls; all 10 returned "No data found"; 0 returned data.**
- The final answer nonetheless stated **Error Rate 0% · P95 245ms · P99 412ms · Throughput 12.4 req/s**, attributed to `container_http_requests_duration_seconds_total` — a metric that returned nothing on every query.

Contrast: a different run on the identical query did **not** fabricate — it found the real throughput (~0.043 rps) and error rate (0%) via discovery, and **honestly reported P95/P99 as unavailable** ("the required duration bucket metrics did not return data"). So the honest path exists; fabrication is reachable but not constant — a **provenance gap in the answer-synthesis (solver) step**, where the model is allowed to emit metric values that aren't grounded in any tool observation.

## Proposed fix (unconditional — not behind any feature flag)

- In the solver / answer-synthesis path, enforce a provenance rule: a numeric metric value MUST be traceable to a tool result in the scratchpad. If the relevant queries returned no data, the answer must state the metric is **unavailable** (and why), and MUST NOT invent a value.
- Prefer enforcing this where answers are assembled/critiqued (e.g. `llm/llm-server/agents/core/planner_rewoo_solver.go` + the provenance check in `llm/llm-server/agents/prompts_repo/prompt_rewoo_critiquer.txt`), so it holds for every agent, independent of any RCA-maturity flag.

## Acceptance criteria

- Given a metrics question where all queries return no data, the answer reports "unavailable" (with the empty-result citation) and contains no fabricated values.
- The guard is always on (no flag) and applies across agents.
- A regression check (eval fixture or test) covering "all-empty queries → no invented numbers".

Related: a separate metric-discovery determinism effort reduces how often queries come back empty, but does not by itself prevent fabrication — this guard is needed regardless.

VaishnaviKumar0704:
Closes #291

## Summary
When the AI assistant's answer-synthesis step encounters metric queries that
returned no data, it could sometimes fabricate plausible-looking numeric values
(e.g. P95/P99 latency, throughput) instead of stating the metric was
unavailable. This adds an unconditional provenance guard so numeric values in
final answers must trace back to an actual tool observation.

## Changes
- **Solver prompts** (`planner_rewoo_solver.txt`, and the `v1`/`v2` active
  variants under `prompts/default/.../rewoo_solver.txt`): added a
  **Numeric Provenance Rule** requiring every stated number to be grounded in
  an observation, with an explicit "unavailable" path when queries return
  empty/error instead of permitting a substituted value.
- **Critiquer prompts** (`prompt_rewoo_critiquer.txt` for the ReWOO path,
  `planner_react_critiquer.txt` for the ReAct path): added a
  **Numeric Provenance Check** rule that rejects any final answer containing a
  number with no traceable source in the scratchpad — this is the
  enforcement/rejection gate, independent of the solver doing the right thing
  on its own.
- **`planner_rewoo_solver.go`**: added a detection-only regex
  (`rewooSolverMetricLikeRegex`) that logs a warning when the solver's
  existing `insufficientData` flag is true and the final answer still
  contains metric-like numeric patterns. This does **not** block or alter the
  answer — exact value-matching against the scratchpad is unreliable
  (rounding, derived rates, unit conversions), so the real guard is the
  prompt-level rules above. This is purely for observability into how often
  the model still attempts this pattern post-fix.

## Tests
Added `planner_rewoo_solver_test.go` covering:
- The detection regex correctly matches metric-like values (ms, %, req/s,
  MB/GB/KB, etc.) and correctly ignores plain numbers / already-flagged
  "unavailable" language
- The `truncateForLog` log-safety helper

## Notes
This guard is applied unconditionally (no feature flag), per the issue's
acceptance criteria, and covers both the ReWOO and ReAct agent critique paths
since the bug wasn't scoped to one specific loop.


api-server/services/observability/splunk_logs.go
```diff
@@ -9,8 +9,39 @@ import (
 	"time"
 )
 
+// o11yLogSearcher abstracts the Splunk O11y log search call so it can be
+// replaced with a fake in unit tests without touching the integrations package.
+type o11yLogSearcher interface {
```
hsundar-git, line 14: high

**This file (and its new test) are unrelated to the PR's stated purpose (#291 metric fabrication).** The diff bundles a substantial Splunk Log-Observer refactor — a new `o11yLogSearcher` test-seam interface and, more importantly, a **behavioral change** to `QueryLabels`: it switches from returning a static field list to **dynamically sampling 200 recent log entries** on every call. That's a real functional + performance change to Splunk label discovery that has nothing to do with preventing fabricated metric values.

The PR also deletes two frontend files (`app/src/reportWebVitals.ts`, `app/src/react-app-env.d.ts`) that are likewise unrelated.

This looks like the branch was cut from a dirty base / picked up someone else's work. Please **rescope the PR to only the #291 metric-provenance change** (the solver/critiquer prompts + `planner_rewoo_solver.go` detection regex + its test) and drop the Splunk and frontend changes — they need their own PR, description, and review (the dynamic-discovery change in particular deserves its own justification and a perf look at sampling 200 entries per `QueryLabels` call).

api-server/services/observability/splunk_logs.go
```diff
@@ -339,9 +398,8 @@ func buildO11yOperatorClause(field string, op query.BinaryWhereClauseType, val a
 		}
 		return fmt.Sprintf("NOT %s:%s", safeField, strVal), nil
 	case query.Like:
-		// Treat like as a wildcard prefix match
 		return fmt.Sprintf("%s:%s*", safeField, integrations.EscapeO11yQueryString(fmt.Sprintf("%v", val))), nil
 	default:
 		return fmt.Sprintf("%s:%s", safeField, strVal), nil
 	}
-}
+}
```
hsundar-git, line 405: medium

This file has **no trailing newline** (the `\ No newline at end of file` in the diff). `gofmt` always terminates files with a newline, so `gofmt -l` will flag this and the `make fmt` / CI gofmt gate will fail. `splunk_logs_test.go` has the same issue. (Both go away if the Splunk changes are split out per the other comment, but flagging in case they stay.)

llm/llm-server/agents/core/planner_rewoo_solver.go
```diff
@@ -422,3 +448,12 @@ func hasToolFailureMajority(scratchpad string) bool {
...
 	return (noData+success) > 0 && (success == 0 || noData > success)
 }
+
+// truncateForLog trims s to at most n runes, appending "..." if truncated. Used to avoid
+// dumping full answer content into structured logs.
+func truncateForLog(s string, n int) string {
+	if len(s) <= n {
+		return s
+	}
+	return s[:n] + "..."
```
hsundar-git, line 458: low

The doc comment says "trims s to at most n **runes**", but `len(s)` and `s[:n]` operate on **bytes** — for multi-byte UTF-8 content (non-ASCII log text) this can slice mid-rune and emit invalid UTF-8 into the structured log. Either fix the comment to say "bytes", or make it rune-safe, e.g.:

```go
r := []rune(s)
if len(r) <= n {
    return s
}
return string(r[:n]) + "..."
```

Log-only, so low severity — but the comment/behavior mismatch is worth resolving.

hsundar-git:
## Summary of Changes

Adds a guard against the AI assistant **fabricating metric values** when a query returns no data (#291). The fix is prompt-led: a "Numeric Provenance Rule" in the ReWOO solver prompts and a "Numeric Provenance Check" rejection gate in both the ReWOO and ReAct critiquer prompts, requiring every stated number to trace back to a real tool observation (with an explicit "unavailable" path for empty/error results). A detection-only regex in `planner_rewoo_solver.go` logs a warning when an `insufficientData` answer still contains metric-like values — observability only, it does not block.

### Review Summary

**Services affected:** llm-server (Go + prompts) — *plus unintended* api-server/observability (Go) and frontend (TS)
**Risk level:** Medium — the intended change is sound, but the PR bundles unrelated changes that must be split out.

|---|---|---|
| Scope / Process | Diff includes unrelated work — a Splunk `QueryLabels` dynamic-discovery refactor (`splunk_logs.go` +88/-30, `splunk_logs_test.go` +203) and frontend deletions (`reportWebVitals.ts`, `react-app-env.d.ts`) — none mentioned in the description (#291) | high |
| Style / CI | `splunk_logs.go` & `splunk_logs_test.go` have no trailing newline → `gofmt`/CI fmt gate fails — `splunk_logs.go:405` | medium |
| Correctness | Frontend deletions look accidental and may break the CRA build: `react-app-env.d.ts` provides the `react-scripts` type reference, and `reportWebVitals.ts` is typically imported by `index.tsx` — deleting them unexplained is risky | medium |
| Correctness | `truncateForLog` slices by **bytes** though its comment says "runes" — can split multi-byte UTF-8 in logs — `planner_rewoo_solver.go:458` | low |
| Coverage | Detection regex omits common units — seconds (`245s`/`1.2s`), `ns`/`µs`, `TB`, and Kubernetes `Ki`/`Mi`/`Gi` — acceptable since it's detection-only, but easy to broaden — `planner_rewoo_solver.go:31` | low |

> ✅ **Looks Good** — the #291 fix itself
>
> The metric-provenance change is well designed: the rule is applied in **all three** relevant prompts (ReWOO solver + the `prompts/default/v1` and `v2` active variants) **and** enforced as a rejection gate in **both** critiquer paths (ReWOO + ReAct) — so it doesn't depend on the solver self-policing. The "grounded vs. unavailable, never substitute a number" framing is clear and explicitly overrides the "complete-looking answer" instinct and the fallback-commands constraint. The Go change is appropriately **detection-only** (the author correctly notes exact value-matching against the scratchpad is infeasible), with unit tests for the regex and `truncateForLog`. Applied unconditionally per the issue's acceptance criteria. Nicely scoped — *if* the unrelated Splunk/frontend changes are removed.

### Checklist
- [x] Tests cover new/changed behavior (regex + truncate; Splunk tests too)
- [x] No hardcoded secrets
- [x] Error handling appropriate (detection-only, never alters the answer)
- [ ] **PR scoped to a single concern** — split out the Splunk + frontend changes (High)
- [ ] `gofmt` clean (trailing newline on the Go files) (Medium)
- [ ] Confirm the frontend deletions don't break the build (Medium)

---
*Automated PR Review*

hsundar-git:
@VaishnaviKumar0704 please resolve the conflicts also provide an information on how this change is being tested and verified as it involves LLM prompt changes?

PR #492, opened 2026-06-28
feat(ui): add Switch Case auto-suggest dropdown and mismatch warnings (#283)

### LINKED ISSUE #283: feat(workflow): Switch Case action — auto-suggest values from previous action + validate before run
## Summary

The **Switch Case** action requires users to manually type the values they're matching against. There's no awareness of what values the upstream action could actually produce, so a small typo (or a case mismatch) silently causes the case to never match — workflow execution fails or takes an unexpected branch with no obvious cause.

The classic example: an upstream Approval Task is configured with options `Yes` / `No`. The downstream Switch Case is then configured with `yes` / `no` (lowercased by habit). The comparison is case-sensitive, no case matches, and the workflow fails — without any pre-run warning that the values don't line up.

## Why it matters

Hand-typing magic string values that have to match another action's output is fragile and exactly the kind of error UX should catch. Three small UX additions would cover almost all cases:

1. **Auto-suggest from the upstream action's output.** When the upstream task has a known finite set of outputs (Approval options, enum-shaped Get-Value results, etc.), the Switch Case value field should offer them as a dropdown / autocomplete.
2. **Validation warning at save time.** If the configured case value doesn't appear in the upstream's known outputs, show an inline warning ("Configured value does not match any available output from the previous action") — without blocking save, because dynamic outputs are also legitimate.
3. **Helper text under the field.** Display the upstream action's known outputs as guidance: *"Available values from previous action: Yes, No."*

## Expected behaviour

In the workflow builder Switch Case configuration panel:

- The "case value" input is augmented with a dropdown of the upstream action's known outputs (where determinable).
- Free-text remains supported (for cases where the upstream produces dynamic values).
- A non-matching value shows an inline warning but does not block save.
- Helper text under the field lists known upstream outputs.

## Reproduction (current behaviour)

1. Add an Approval Task with options `Yes` / `No`.
2. Add a Switch Case downstream and type the cases as `yes` and `no` (lowercase).
3. Run the workflow → no case matches → branch taken is undefined / workflow fails.
4. No pre-run warning that the values can't possibly match.

## Where to start (rough orientation for a contributor)

- **Workflow builder UI (`app/`):** find the Switch Case action's configuration panel. The value input is currently a plain text field — it needs to become an autocomplete-style input.
- **Upstream-output discovery:** the workflow builder already maintains a graph of tasks (the canvas). For each task type with a known finite output set (Approval, enum-shaped Get-Value, etc.), expose that set so the Switch Case panel can read it.
- **Validation:** a save-time check that the configured value appears in the upstream's known outputs (when the upstream's outputs are determinable). Warn-only, don't block.

(Verify the exact paths in your checkout — these are rough pointers.)

## Acceptance

After the change:
1. Adding a Switch Case downstream of an Approval Task shows the approval's options as autocomplete suggestions for each case value.
2. Typing a value that doesn't match any upstream output shows an inline warning (but save still works — dynamic values are valid).
3. Helper text under the case-value input lists the known upstream outputs.
4. Existing workflows continue to work — the change is additive to the editor; no migration required.

## Out of scope

- Changing the comparison from case-sensitive to case-insensitive (deliberately rejected — the comparison semantics are part of the workflow contract; the UX should help users *type* the right value, not silently rewrite values at runtime).

reckless-sherixx:
Resolves https://github.com/nudgebee/nudgebee/issues/283

# Walkthrough - Switch Case Auto-Suggest from Upstream Action Outputs

I have implemented an intelligent auto-suggest, save-time warning, and helper text feature for Switch Case actions (`core.switch`) when matching against upstream outputs with known finite values (such as `core.approval`).

### 1. Upstream Field Configuration (Creation of Options)
When configuring an upstream task such as an **Approval Action** (`core.approval`), users define finite output values in the **Approval Options** field (e.g. adding items like `Yes` and `No` via the array editor). These defined options establish the exact contract of values that the action can produce during workflow execution.
{D7C5CDD5-F672-4FCE-AFC2-D7B8AB736495}

### 2. Downstream Auto-Suggestion in Switch Cases
When adding a downstream **Switch Case Action** (`core.switch`) configured to evaluate an expression from that upstream task (e.g. `{{ Tasks['approval_1'].output.status }}`), the UI now automatically recognizes the upstream field creation and provides:
- **Autocomplete Dropdown Suggestions**: Clicking inside the Case Value input field presents a dropdown list containing the exact options configured in the upstream action (`Yes`, `No`). Users can pick from the list or type freely (`freeSolo`).
- **Helper Text Guidance**: Informative descriptive text below the field displays the known available values (`Available values from approval_1: Yes, No`).
- **Inline Mismatch Warning**: If a user hand-types a value that does not match any known upstream output (e.g. typing `yes` when the option is cased as `Yes`, or a typo like `invalid_option`), a non-blocking amber warning banner appears right below the field (`⚠ This value does not match any known output from the upstream action.`).
---
#### `getUpstreamSuggestedValues.ts`
- Created a pure function that parses the switch case `expression` (supporting single quotes `Tasks['id']` and double quotes `Tasks["id"]`).
- Locates the referenced node in the graph and extracts known finite output values (e.g. `approval_options` from `core.approval` nodes, defaulting to `["approve", "reject"]`).
- Detects piped case-conversion filters (`| lower`, `| upper`) and transforms suggested values appropriately.
- Detects multi-task expressions and returns an `isComplexExpression` flag.

#### `WorkflowFieldComponents.tsx`
- Augmented `ArrayEditor` to accept `suggestedValues`, `upstreamTaskInfo`, and `isComplexExpression` props.
- When rendering switch case `value` inputs, displays a MUI `<Autocomplete freeSolo>` when suggestions are available, allowing user selection while preserving custom free-text entry.
- Displays helpful descriptive guidance under the field listing known upstream outputs or noting complex expression states.
- Displays an inline warning (`⚠ This value does not match any known output from the upstream action.`) in amber when an entered value does not match any known suggestion.

#### `ActionDetailsSidebar.tsx`
- Memoized `switchUpstreamValues` computation using the selected switch node and graph state.
- Threaded suggested values props into `<ArrayEditor>` specifically when editing switch `cases`.

### Browser UI Verification
We tested the implementation end-to-end in the Nudgebee Workflow Builder (`http://localhost:3000/workflow/new`) by injecting a test workflow with an upstream Approval task (`approval_1`) connected to our Switch Case action (`switch_1`):
1. **Helper Text Guidance**: Below the case value field, helper text explicitly shows the available finite values from the upstream task (`Available values from Approval: Yes, No`).
{FB112B27-64E6-4B68-9499-B399577C21F1}

2. **Autocomplete Dropdown Suggestions**: Focusing or clicking inside the Case Value field presents an Autocomplete dropdown with the valid options (`Yes` and `No`).
{68C0DA7A-CE7C-4F6C-8DD8-4D28501F7A82}

3. **Inline Mismatch Warning Banner**: Typing a value that does not match any known upstream output (`invalid_option`) triggers a non-blocking amber warning banner right below the input.
{19F13644-28ED-443C-AAF9-E28845DF5080}

4. **

### Automated Tests
- Created `getUpstreamSuggestedValues.test.ts` covering single/double quoted references, defaults, multi-task expressions, and piped filters.
- Executed `npx jest getUpstreamSuggestedValues.test.ts` — **All 9 tests PASSED**.
- Executed `npm run type-check` — **PASSED**.
- Executed `npm run lint2` — **PASSED**.


app/src/components1/workflow/utils/getUpstreamSuggestedValues.ts
```diff
@@ -0,0 +1,93 @@
...
+  const taskRefRegex = /Tasks\[['"]([^'"]+)['"]\]\.output\.([a-zA-Z0-9_]+)/g;
+  const matches = Array.from(expression.matchAll(taskRefRegex));
+
+  if (matches.length === 0) {
+    return null;
+  }
+
+  // Check if there are multiple distinct task references
+  const distinctRefs = new Set(matches.map((m) => `${m[1]}.${m[2]}`));
+  if (distinctRefs.size > 1) {
```
Kankshit-02: high

**Single-ref guard is too loose — false suggestions + false mismatch warnings.**

The switch matches on *exact post-render string equality* (`switch_task.go`: `if expressionValue != caseValue { continue }`). This guard only rejects expressions with **multiple distinct `Tasks[...]` refs**. It still returns suggestions when the expression is a single task ref surrounded by literal text or mixed with other namespaces, e.g.:

- `prefix-{{ Tasks['a'].output.status }}` -> renders `prefix-Yes`, but you suggest `Yes`
- `{{ Tasks['a'].output.status }}-{{ Inputs.env }}` -> one Tasks ref, `distinctRefs.size === 1`, suggestions returned

In both cases the rendered value can never equal a raw suggestion, so the mismatch `⚠` warning fires on a **correct** config. Suggestions are only sound when the expression is *solely* the single task ref (optionally with a whitelisted filter). Recommend: strip the matched ref + whitespace + optional `| filter` and confirm nothing else remains; otherwise treat as complex.

app/src/components/workflow/components/WorkflowFieldComponents.tsx
```diff
@@ -11,6 +11,7 @@ import {
   FormControlLabel,
   ToggleButtonGroup,
   ToggleButton,
+  Autocomplete,
```
Kankshit-02: We can use components from ds instead of using autocomplete

Kankshit-02: You can also replace the rest of the mui imports with ds either as a part of this pr or another

Kankshit-02 (approved):

PR #506, opened 2026-06-29
fix: prevent AI assistant from fabricating metric values on empty query results

### LINKED ISSUE #291: [BUG] - AI assistant sometimes reports made-up metric values when the underlying query returned no data

VaishnaviKumar0704:
Closes #291

## Summary
When metric queries return no data, the AI assistant's answer-synthesis step
could fabricate plausible-looking numbers (e.g. P95/P99 latency, throughput)
instead of reporting the metric as unavailable. This adds an unconditional
provenance guard across all solver and critiquer prompt paths.

## Changes
- **Solver prompts** (`planner_rewoo_solver.txt`, `v1/rewoo_solver.txt`,
  `v2/rewoo_solver.txt`): added Numeric Provenance Rule — every stated number
  must be grounded in an observation; explicit "unavailable" path required when
  queries return empty/error.
- **Critiquer prompts** (`prompt_rewoo_critiquer.txt`,
  `planner_react_critiquer.txt`): added Numeric Provenance Check rule that
  rejects any final answer with an ungrounded number — covers both ReWOO and
  ReAct agent paths.
- **`planner_rewoo_solver.go`**: detection-only warning log when
  `insufficientData` is true but the answer contains metric-like values.
  Does not block answers — exact scratchpad matching is unreliable due to
  rounding/derived values; real enforcement is in the prompts.

## Tests
- `planner_rewoo_solver_test.go`: regex matching tests + `truncateForLog` tests
  including multi-byte UTF-8 and emoji edge cases (rune-safe implementation).

## Notes
No feature flag — guard is always on, per acceptance criteria in #291.


llm/llm-server/agents/core/planner_rewoo_solver.go
```diff
@@ -422,3 +448,15 @@ func hasToolFailureMajority(scratchpad string) bool {
 	noData := failed + empty
 	return (noData+success) > 0 && (success == 0 || noData > success)
 }
+
+// truncateForLog trims s to at most n runes, appending "..." if truncated. Used to avoid
+// dumping full answer content into structured logs. Operates on runes, not bytes, so it
+// never splits a multi-byte UTF-8 character mid-sequence (log content may include non-ASCII
+// text from logs, error messages, or service names).
+func truncateForLog(s string, n int) string {
```
hsundar-git, line 456: > ✅ **Looks Good**

This directly addresses the byte-vs-rune finding from the prior iteration (#490). `[]rune(s)` slicing is correct and can never split a multi-byte sequence, and the new `TestTruncateForLog` cases (`hëllo`, `abc😀😀😀def`) lock that behavior in. Nice — exactly the rune-safe form.

llm/llm-server/agents/prompts_repo/planner_react_critiquer.txt
```diff
@@ -171,4 +179,4 @@ Tool Details:
 
 <final_answer>
 {{.final_answer}}
-</final_answer>
+</final_answer>
```
hsundar-git, line 182: low

This change drops the trailing newline at end of file. The repo's `.editorconfig` declares `insert_final_newline = true` for `[*]`, so this is contrary to the stated convention. The same trailing-newline removal appears in `planner_rewoo_solver.txt` and both `prompts/default/v1|v2/utilities/rewoo_solver.txt`.

It won't fail CI (`.pre-commit-config.yaml` only runs black/flake8/mypy — no `end-of-file-fixer`), so this is purely a nit — worth restoring the final newline on all four prompt files to stay editorconfig-clean. The unrelated de-indentation of the ` ```xml ` fences in these same files is also cosmetic churn outside the #291 scope, though harmless for the embedded prompt text.

hsundar-git:
## Summary of Changes

Prevents the AI assistant from fabricating metric values (P95/P99 latency, throughput, error rates, etc.) when the underlying queries return no data. Adds an always-on **Numeric Provenance** guard across every solver and critiquer prompt path, plus a detection-only warning log in the ReWOO solver. This is the re-scoped **v2** of #490 — Closes #291.

### Highlights

- **Solver prompts** (`planner_rewoo_solver.txt`, `v1`/`v2` `rewoo_solver.txt`): "Numeric Provenance Rule" — every stated number must be grounded in an observation or explicitly flagged unavailable; explicitly noted as *not* overridden by the CLI-fallback rule.
- **Critiquer prompts** (`prompt_rewoo_critiquer.txt`, `planner_react_critiquer.txt`): "Numeric Provenance Check" that `refine`/REJECTs any final answer containing an ungrounded number — covers both ReWOO and ReAct paths, so it doesn't rely on the solver self-policing.
- **`planner_rewoo_solver.go`**: detection-only `Warn` log when `insufficientData` is true but the answer contains metric-like values. Correctly does **not** block (exact value-matching against the scratchpad is infeasible), with `truncateForLog` + regex unit tests.

### Review Summary

**Services affected:** llm-server (Go + prompts)
**Risk level:** Low

**Addresses both substantive findings from #490:**
- ✅ Scope creep removed — the unrelated Splunk `QueryLabels` refactor and frontend deletions are gone; the diff is now confined to the #291 fix.
- ✅ `truncateForLog` is now rune-based (`[]rune` slicing), with explicit multi-byte/emoji tests — the byte-vs-rune bug is fixed and locked in.

Verified during review: the detection log is gated on `insufficientData` (observability only, never blocks); the `%` alternative is correctly split out of the `\b`-terminated group (a `\b` won't match after `%`) and is covered by tests; dual solver+critiquer enforcement across both v1/v2 prompt variants means a single self-policing failure can't leak a fabricated number.

|---|---|---|
| Correctness | Rune-safe `truncateForLog` + regex, both unit-tested — core fix is sound | > ✅ |
| Style | Trailing newline removed from 4 prompt `.txt` files, contrary to `.editorconfig` `insert_final_newline = true` (not CI-enforced) — `planner_react_critiquer.txt:182` and the 3 solver prompts | low |
| Scope | Unrelated ` ```xml ` fence de-indentation churn in the same prompt files (cosmetic, harmless for embedded prompts) | low |
| Coverage | Detection regex omits some units (`s`/`ns`/`µs`, `KiB`/`MiB`/`GiB`, durations `m`/`h`) — acceptable since it's detection-only and disclosed, but worth noting | low |

### Checklist
- [x] Tests cover new/changed behavior (regex + `truncateForLog`, incl. UTF-8/emoji)
- [x] No hardcoded secrets
- [x] Error handling is appropriate (detection log is non-blocking by design)
- [x] No breaking API changes
- [x] Linting/formatting passes for affected services (Go file has trailing newline; gofmt-clean)

Clean, well-scoped iteration that fully resolves the #490 feedback. The dual prompt-level enforcement (solver + both critiquers, v1 + v2) is the right design — the Go side stays observability-only where exact matching can't be trusted. LGTM; the remaining notes are all cosmetic nits.

---
*Automated PR Review*

hsundar-git:
@VaishnaviKumar0704 heads up for you on the orchestrators we have - we are going to remove rewoo and react2 and will be moving all to react3 - React3 will be single new ract orchestrator - Please update the PR accordingly 
- Also please provide more information on how it's been tested against an actual data with real llm provider 
- You can reach out to me/team on discord for help for setting up some of these things

PR #539, opened 2026-07-02
security(collector,notifications): clear OS + cryptography CVEs from stale python base

mayankpande88:
# Description

k8s-collector and notifications-server both build on `ghcr.io/nudgebee/nudgebee-python:3.12-20250919-140344` — a base last rebuilt **Sep 2025**, whose Alpine OS packages have drifted behind security patches (openssl/`libcrypto3`/`libssl3`, `musl`, `zlib`, `nghttp2`).

## Changes

| Service | Fix | Result |
|---|---|---|
| k8s-collector | Add `apk upgrade --no-cache` in the production stage | **2 CRIT + 19 HIGH → 0/0** |
| notifications-server | Already runs `apk upgrade` (OS clears on rebuild); bump `cryptography` 48.0.0 → 48.0.1 (GHSA-537c-gmf6-5ccf) | **3 HIGH → 0/0** |

Both verified locally with `docker build` + `trivy image` → **CRITICAL 0 / HIGH 0**.

The notifications image installs from `pyproject.toml` via `uv`, so the pin bump fixes the image; `poetry.lock` was regenerated for CI (`poetry install`) consistency — only `cryptography` changed version (the rest of the lock diff is that package's wheel hashes).

## Root cause / follow-up

The real cause is that **`ghcr.io/nudgebee/nudgebee-python` has no automated rebuild/publish** — the infra `python-base-image-deploy*.yaml` workflows push to **ECR** (enterprise), never GHCR, so the OSS base is a stale, manually-pushed artifact. A separate PR adds an OSS base-image → GHCR publish workflow (with a schedule) so the base stays patched and this `apk upgrade` becomes a cheap no-op. Same gap underlies the cloud-collector/llm/ml-k8s OS CVEs.

## Type of change

- [x] Security (non-breaking change which improves security)

# How Has This Been Tested?

- [x] `docker build` + `trivy image` for both services → CRIT 0 / HIGH 0
- [x] `poetry lock` regenerated; only cryptography version changed

PR #540, opened 2026-07-02
ci(base): publish nudgebee-python base image to GHCR (weekly)

mayankpande88:
# Description

Durable root-cause fix for the stale OSS Python base (see #539 for the immediate consumer-side fix).

## Problem

`ghcr.io/nudgebee/nudgebee-python` — which **k8s-collector** and **notifications-server** build `FROM` — has **no publisher**. The enterprise infra workflows (`python-base-image-deploy*.yaml`) push this base only to **ECR**, so the public GHCR tag was pushed once (`3.12-20250919-140344`, Sept 2025) and never refreshed. Its Alpine OS packages drifted behind security patches and consumers inherited the CVEs (openssl/musl/zlib/nghttp2).

## Change

Add an OSS workflow that builds `deploy/kubernetes/nudgebee-python-base-image` and publishes it to GHCR — **dual-arch (amd64 + arm64, native runners) with a manifest merge**, mirroring `edge.yaml` / `nudgebee-debug.yaml`.

Tags:
- `:3.12` — rolling (consumers pin this for auto-fresh patches)
- `:3.12-<YYYYMMDD>-<sha7>` — immutable snapshot for reproducible pins

**Freshness:** the base Dockerfile already runs `apk upgrade`, so every rebuild picks up the latest Alpine patches. A **weekly `schedule`** rebuilds even with no code change, so the base can't silently go stale again. Verified locally: a fresh rebuild ships **openssl 3.5.7-r0** (the fix for the currently-flagged CVEs).

Triggers: push to the base dir / this workflow, weekly cron, manual dispatch; PRs build both arches to validate without pushing.

## Follow-ups

1. After this publishes `:3.12`, bump `PYTHON_BASE` in k8s-collector + notifications-server to it (then the `apk upgrade` in #539 becomes a cheap no-op / can be dropped).
2. Same pattern should cover the `nudgebee-ml-base` and `nudgebee-cloud-collector-base` images (behind the cloud-collector/llm/ml-k8s OS CVEs).
3. **First push:** flip the new `ghcr.io/nudgebee/nudgebee-python` package to public in Settings → Packages (new GHCR packages default to private).

## Type of change

# How Has This Been Tested?

- [x] Workflow YAML validated
- [x] Base image builds locally (dual-stage) → Python 3.12.13, openssl 3.5.7-r0
- [x] Structure mirrors the working `nudgebee-debug.yaml` (prepare → per-arch build → manifest merge)


mayankpande88:
I have read the CLA Document and I hereby sign the CLA

blue4209211:
I have read the CLA Document and I hereby sign the CLA

blue4209211 (approved):

Issue #497, opened 2026-06-28
OpenObserve: log/metric/trace query sources


blue4209211:
Parent epic: #494

## Scope

Implement the three observability query sources for OpenObserve under `api-server/services/observability/`:

- `openobserve_logs.go` — `LogSource` over `POST /api/{org_id}/_search` (SQL + `start_time`/`end_time` in microseconds).
- `openobserve_metrics.go` — `MetricSource` via SQL and PromQL endpoints.
- `openobserve_traces.go` — `TraceSource` over `/api/{org_id}/{stream}/traces/latest` and the search API.

Interfaces: `LogSource` / `TraceSource` / `MetricSource` from `observability/entity.go` + `service.go`.

Reference: `api-server/services/observability/datadog.go`, `newrelic_logs.go`, `newrelic_metrics.go`, `newrelic_traces.go`.

## Definition of done

- All three sources implemented and registered.
- Time-range, query, and pagination parameters are honoured.
- Errors from OpenObserve are surfaced verbatim to the caller.
- Unit tests with a fake HTTP server cover the response shapes.

blue4209211:
Closing as completed on current main. OpenObserve log, metric, and trace sources plus their tests are present under ; the implementation originally landed via PR #567 and has received subsequent fixes.

blue4209211:
Triage clarification: current main contains the OpenObserve log, metric, and trace sources plus their tests under api-server/services/observability. The foundation landed through PR 567 and has received subsequent fixes.

Issue #495, opened 2026-06-28
OpenObserve: DB migration to register integration type


blue4209211:
Parent epic: #494

## Scope

Add an `integration_types` seed row registering **OpenObserve** under category `observability_platform`, mirroring the Datadog/NewRelic seeds.

Reference: `api-server/migrations/migrations/app/1751348269124_V541_insert_into_public_integration_types.up.sql`

## Definition of done

- New `up.sql` / `down.sql` migration created via `./api-server/migrations/new-migration.sh add_openobserve_integration_type`.
- Row inserts under category `observability_platform` with provider name `openobserve`.
- Migration is idempotent (`ON CONFLICT DO NOTHING` or equivalent).
- `make validate` passes in `api-server/services`.

blue4209211:
Closing as completed on current main. Migration  registers  under  with ; the matching down migration and updated  are present.

blue4209211:
Triage clarification: current main contains migration 1784695611998_V774_add_openobserve_integration_type.up.sql, registering openobserve under observability_platform idempotently, plus the matching down migration and updated atlas.sum.

PR #567, opened 2026-07-04
feat: Implement OpenObserve Logs, Metrics, and Traces integrations

reckless-sherixx:
### Description
This PR implements the OpenObserve observability query sources, enabling OpenObserve as a supported provider for Logs, Metrics, and Traces in Nudgebee.

**Fixes:** #497 

### Changes Made

- **Source Registration (`observability/service.go`)**: 
  - Registered `OpenObserveLogSource`, `OpenObserveMetricSource`, and `OpenObserveTraceSource` in the central `service.go` factory.
  - Added capability mappings for OpenObserve in `allProviderCaps` (`SupportsServiceMap: true`, `SupportsRawQuery: true`, etc.).
- **Logs Source (`observability/openobserve_logs.go`)**:
  - Implemented the `LogSource` interface.
  - Queries `POST /api/{org_id}/_search` using an SQL-like query constructed by parsing Nudgebee's `QueryWhereClause`.
  - Handles parsing OpenObserve JSON log responses directly into Nudgebee's generic `OpenTelemetryLog` objects.
- **Metrics Source (`observability/openobserve_metrics.go`)**:
  - Implemented the `MetricSource` interface.
  - Queries `POST /api/{org_id}/prometheus/api/v1/query_range` supporting the Prometheus spec.
  - Passes through native Prometheus `matrix` format responses for seamless upstream compatibility.
- **Traces Source (`observability/openobserve_traces.go`)**:
  - Implemented the `TraceSource` interface.
  - Translates Nudgebee generic `QueryWhereClause` AST into OpenObserve compatible SQL statements.
  - Maps metadata (like `trace.id`, `duration.ms`, `span.id`) smoothly into Nudgebee's generic `OpenTelemetryTrace` model.
- **Unit Testing**:
  - Added robust unit testing suites (`openobserve_logs_test.go`, `openobserve_metrics_test.go`, `openobserve_traces_test.go`) utilizing mocked HTTP servers. Test suites ensure SQL AST builds correctly and response shapes are accurately parsed.

### Verification
- [x] All three observability sources implemented and registered.
- [x] Time-range, query operators, and pagination parameters are accurately translated to OpenObserve semantics.
- [x] Tested against all scenarios in mocked `httptest.NewServer` environments.
- [x] Build passes (`go build ./...`)
- [x] All Unit tests pass (`go test -v ./observability -run TestOpenObserve`)


mangglesh:
### Missing migration: `openobserve` row in `integration_types`

Creating an OpenObserve integration fails at insert time:

```
failed to create integration: pq: insert or update on table "integrations"
violates foreign key constraint "integrations_type_fkey" (23503)
```

**Root cause:** `integrations.type` is an FK to `integration_types(name)`. This PR adds the Go source, the `service.go` registration, and the UI wiring, but never seeds the `openobserve` lookup row — so every insert of `type='openobserve'` is rejected. Every other provider ships this row via a migration (e.g. `V690_solarwinds_integration`, and the `observe`/`loki`/`signoz` seeds).

**Fix:** add a migration seeding the row, matching the SolarWinds pattern:

```sql
-- <ts>_V772_openobserve_integration_type.up.sql
INSERT INTO integration_types(name, category, description) VALUES
  ('openobserve', 'observability_platform', 'OpenObserve Observability Platform')
ON CONFLICT (name) DO NOTHING;

-- .down.sql
DELETE FROM integration_types WHERE name = 'openobserve';
```

Use `./api-server/migrations/new-migration.sh openobserve_integration_type` to scaffold with a correct timestamp/version rather than hand-picking one.

---

Separately, while validating the integration locally I hit a red herring worth noting for testers: the `TestConnection` "Test Credentials" step returns `Get "…/api/{org}/streams": EOF` when OpenObserve isn't actually reachable at the configured URL (e.g. a dead `kubectl port-forward` — the local port accepts the TCP connection but the tunnel is broken). Not a code issue, but the error is opaque; may be worth surfacing a clearer hint on `EOF`/connection errors.

reckless-sherixx:
@mangglesh this pr focuses on fixing issue #497 and the changes you are suggesting for the migration are from issue #495.  So should i fix the error message for this pr or fix both issues in this pr only.

mangglesh:
> @mangglesh this pr focuses on fixing issue #497 and the changes you are suggesting for the migration are from issue #495. So should i fix the error message for this pr or fix both issues in this pr only.

yes please fix both issues in this PR itself. so that this PR can become complete feature end to end.

reckless-sherixx:
@mangglesh Made the changes

mangglesh:
> @mangglesh Made the changes

checking on local

api-server/services/observability/openobserve_logs.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+// OpenObserveLogSource implements LogSource interface for OpenObserve
+type OpenObserveLogSource struct{}
+
+// openObserveLogLabelMapping maps standard field names to OpenObserve field names
+var openObserveLogLabelMapping = map[string]string{
+	"timestamp": "_timestamp",
+	"body":      "message",
+	"message":   "message",
```
mangglesh: **Bug: log field mapping points at a non-existent column.** OTel-ingested logs store the log line in `body` — there is no `message` field. Mapping `body`/`message` → `message` makes every text filter build `WHERE ... message ...`, which OpenObserve rejects:

```
HTTP 400 {"code":20004,"message":"Search field not found: Schema error: No field named message. Valid fields are default._timestamp, default.body, ..."}
```

Suggested fix:
```go
"body":    "body",
"message": "body",
```
The response parser already reads `body` into `Message` (line ~253), so only the query-side mapping is wrong.

mangglesh (commented):
Tested the three OpenObserve sources end-to-end against a live instance (OTel-ingested logs, `default` stream, ~666K docs). Data flows and `_eq`/`_neq` + And/Or/Not composition are correct, but I found three correctness bugs that make log search return wrong/empty results. Inline comments below with reproductions and suggested fixes. Happy to push a commit with all of them + updated unit tests (the existing test assertions bake in the buggy SQL).

api-server/services/observability/openobserve_logs.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+		for op, val := range ops {
+			strVal := fmt.Sprintf("%v", val)
+			switch op {
+			case query.Eq:
+				parts = append(parts, fmt.Sprintf("%s = '%s'", col, escapeOpenObserveString(strVal)))
+			case query.Nq:
+				parts = append(parts, fmt.Sprintf("%s != '%s'", col, escapeOpenObserveString(strVal)))
+			case query.Contains:
+				parts = append(parts, fmt.Sprintf("str_match(%s, '.*%s.*')", col, escapeOpenObserveString(strVal)))
```
mangglesh: **Bug: `str_match` is a substring match, not regex — the `.*...*` wrapper searches for literal characters.** This affects both `_contains` (here) and `_ilike` (line 89), and since traces reuse this builder, trace filters too.

Measured against live data (last 6h), searching for `error`:

| query | matches |
|---|---|
| `body LIKE '%error%'` (ground truth) | **643,921** |
| `str_match(body, 'error')` | **643,921** ✅ |
| `str_match(body, '.*error.*')` ← this PR | **23** ❌ |
| `re_match(body, '.*error.*')` (actual regex) | 643,921 |

So a UI search for "error" silently returns 23 rows instead of 643,921. Fix — drop the wrapper (`str_match`/`str_match_ignore_case` already mean case-sensitive/insensitive *contains*):
```go
case query.Contains:
    parts = append(parts, fmt.Sprintf("str_match(%s, '%s')", col, escapeOpenObserveString(strVal)))
case query.ILike:
    parts = append(parts, fmt.Sprintf("str_match_ignore_case(%s, '%s')", col, escapeOpenObserveString(strVal)))
```

api-server/services/observability/openobserve_logs.go
```diff
@@ -0,0 +1,374 @@
...
+
+func (s *OpenObserveLogSource) QueryLabels(ctx *security.RequestContext, req FetchLogLabelRequest) ([]OutputLogLabel, error) {
+	// Simple implementation: fetch a single recent log to infer labels from the default stream
+	fetchReq := FetchLogRequest{
+		AccountId:         req.AccountId,
+		LogProvider:       req.LogProvider,
+		LogProviderSource: req.LogProviderSource,
+		StartTime:         req.StartTime,
+		EndTime:           req.EndTime,
+		Limit:             1,
```
mangglesh: **Bug: labels never populate in the UI.** The label-autocomplete call arrives with no time range (`start=end=0`), so sampling logs over `[0,0]` returns nothing — the dropdown is always empty. Two issues:

1. No default window when start/end are 0 → default to e.g. last 1h. (SigNoz avoids this entirely by using a schema-suggest call rather than sampling data.)
2. `Limit: 1` → labels come from a single row, so any field missing on that one log never shows. Sample ~100 rows and union their keys.

Better still: OpenObserve exposes the stream schema (`GET /api/{org}/streams/{stream}/schema`) which returns field names with no data scan — instant, and avoids the timeout on busy streams.

api-server/services/observability/openobserve_metrics.go
```diff
@@ -0,0 +1,281 @@
...
+				continue
+			}
+		}
+
+		endpoint := fmt.Sprintf("%s/api/%s/prometheus/api/v1/query_range", url, orgID)
+
+		form := neturl.Values{}
+		form.Add("query", query)
+		form.Add("start", strconv.FormatInt(req.StartTime, 10))
+		form.Add("end", strconv.FormatInt(req.EndTime, 10))
```
mangglesh: **Bug: time-unit mismatch — `StartTime`/`EndTime` are milliseconds, but they're passed to the Prometheus API as seconds.** The repo convention is that these fields are ms — the native Prometheus source divides by 1000: `time.Unix(req.StartTime/1000, 0)` ([prometheus.go#L458](https://github.com/nudgebee/nudgebee/blob/main/api-server/services/observability/prometheus.go#L458)). Here they're sent raw, so the query window is ~1000× in the future. (A lenient OpenObserve clamps this and still returns series, which hides it.)

```go
form.Add("start", strconv.FormatInt(req.StartTime/1000, 10))
form.Add("end",   strconv.FormatInt(req.EndTime/1000, 10))
```

api-server/services/observability/openobserve_metrics.go
```diff
@@ -0,0 +1,281 @@
...
+		return OutputMetricQuery{}, fmt.Errorf("failed to get OpenObserve configs: %w", err)
+	}
+
+	results := OutputMetricQuery{Results: []QueryResult{}}
+
+	// Calculate step
+	durationSecs := req.EndTime - req.StartTime
+	step := req.StepInterval
+	if step <= 0 {
+		step = int(durationSecs / 100)
```
mangglesh: **Same unit bug in the auto-step.** `durationSecs := req.EndTime - req.StartTime` (line 41) is actually **milliseconds**, so `step = durationSecs / 100` is 1000× too large — e.g. a 1h range yields `step ≈ 36000s` (10h) instead of ~36s, collapsing the series to a single point when the UI doesn't pass `StepInterval`. Fix:
```go
durationSecs := (req.EndTime - req.StartTime) / 1000
```

api-server/services/observability/openobserve_traces.go
```diff
@@ -0,0 +1,370 @@
...
+	searchReq.Query.StartTime = startTimeMicros
+	searchReq.Query.EndTime = endTimeMicros
+
+	payloadBytes, err := json.Marshal(searchReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal search request: %w", err)
+	}
+
+	// For traces, we query the _search endpoint with type=traces if possible, or just the default stream
+	endpoint := fmt.Sprintf("%s/api/%s/_search?type=traces", url, orgID)
```
mangglesh, line 154: **Untested end-to-end.** The live instance has no trace streams (`streams?type=traces` → `[]`), so this path is unverified. Two things to confirm once traces are ingested:
1. The stream name is hardcoded to `"default"` and `?type=traces` — verify OpenObserve returns spans for that combination (the actual traces stream name may differ).
2. The WHERE builder is shared with logs, so `_contains`/`_ilike` here have the same `str_match(..., '.*val.*')` substring bug flagged in the logs review — trace text filters will under-match until that's fixed.

api-server/services/observability/openobserve_traces.go
```diff
@@ -0,0 +1,370 @@
...
+				trace.ParentSpanID = strVal
+			case "name", "span_name":
+				trace.SpanName = strVal
+			case "service.name":
+				trace.ServiceName = strVal
+			case "span_kind":
+				trace.SpanKind = strVal
+			case "duration":
+				if dur, ok := v.(float64); ok {
+					trace.DurationNs = int64(dur)
```
mangglesh: **Likely duration-unit mismatch (please verify with real trace data).** OpenObserve stores span `duration` in **microseconds**, but it's assigned straight into `DurationNs` (nanoseconds) with no conversion, which would make every span duration 1000× too small in the UI. If confirmed:
```go
trace.DurationNs = int64(dur) * 1000
```

mangglesh (commented):
Follow-up: audited the **metrics** and **traces** sources too.

- **Metrics** — list (3,419 metrics), labels (223), and `query_range` all work against a live instance, but there's a time-unit bug (details inline).
- **Traces** — I could not verify against live data: the instance has **no trace streams** (`GET /api/default/streams?type=traces` → `[]`), so `QueryTraces`/`CountTraces` are untested end-to-end. Inline notes cover the static risks (`_contains`/`_ilike` inherit the `str_match` substring bug already flagged in logs, plus a likely duration-unit mismatch).

api-server/services/observability/openobserve_traces.go
```diff
@@ -0,0 +1,370 @@
...
+	"strings"
+	"time"
+)
+
+// OpenObserveTraceSource implements TraceSource interface for OpenObserve
+type OpenObserveTraceSource struct{}
+
+// openObserveTraceLabelMapping maps standard field names to OpenObserve trace field names
+var openObserveTraceLabelMapping = map[string]string{
+	"workload_namespace":        "k8s.namespace.name",
```
mangglesh: **Bug: trace mapping targets columns that don't exist in OpenObserve.** OpenObserve flattens span fields with underscores (dots → `_`); the dotted names here all 400. Verified live against ingested spans:

```
WHERE service.name IS NOT NULL      -> HTTP 400 No field named "service.name"
WHERE otel.status_code IS NOT NULL  -> HTTP 400 No field named "otel.status_code"
WHERE name IS NOT NULL              -> HTTP 400 No field named name
```

Actual span columns: `trace_id, span_id, reference_parent_span_id, operation_name, service_name, duration, span_kind, span_status, status_code, service_k8s_namespace_name, ...`

Corrected mapping (each verified — filters return counts instead of 400):
```go
var openObserveTraceLabelMapping = map[string]string{
    "workload_namespace":        "service_k8s_namespace_name", // was k8s.namespace.name
    "workload_name":             "service_name",               // was service.name
    "span_name":                 "operation_name",             // was name
    "status_code":               "status_code",                // was otel.status_code (numeric 0/1/2; span_status for OK/ERROR/UNSET)
    "parent_id":                 "reference_parent_span_id",   // was parent_id
    "trace_id":                  "trace_id",
    "span_id":                   "span_id",
    // attribute-derived (underscore convention; only present with matching instrumentation):
    "http_status_code":          "http_response_status_code",
    "resource":                  "http_url",
    "destination_workload_name": "service_peer_name",
}
```
Also: identifiers are emitted unquoted and `isSafeIdentifier` allows `.`, so a dotted name like `service.name` is parsed as `table.column` — another reason it fails.

api-server/services/observability/openobserve_traces.go
```diff
@@ -0,0 +1,370 @@
...
+			strVal := fmt.Sprintf("%v", v)
+			switch k {
+			case "trace_id":
+				trace.TraceID = strVal
+			case "span_id":
+				trace.SpanID = strVal
+			case "parent_id", "parent_span_id":
+				trace.ParentSpanID = strVal
+			case "name", "span_name":
+				trace.SpanName = strVal
```
mangglesh, line 205: **Response parser has the same mismatch** — it reads `name`/`span_name` (→ SpanName), `service.name` (→ ServiceName, line 202) and `parent_id` (→ ParentSpanID, line 198), none of which are the real column names, so those fields come back empty in the UI. Use the actual columns:
```go
case "reference_parent_span_id":
    trace.ParentSpanID = strVal
case "operation_name":
    trace.SpanName = strVal
case "service_name":
    trace.ServiceName = strVal
```
Verified: parsing a real span this way yields SpanName=`okey-dokey-0`, ServiceName=`nudgebee-trace-test`, ParentSpanID populated.

api-server/services/observability/openobserve_traces.go
mangglesh: **Duration unit bug — confirmed with live data.** OpenObserve's `duration` is **microseconds**. On a real span: `duration = 123` while `end_time - start_time = 123000` ns → 123 µs. Assigning `int64(dur)` straight into `DurationNs` makes every duration 1000× too small. Fix:
```go
trace.DurationNs = int64(dur) * 1000
```

mangglesh (commented):
Traces are now ingested in the test instance (telemetrygen spans), so I re-ran the trace audit end-to-end. **The trace field mapping does not match OpenObserve's actual span schema** — OpenObserve stores spans with underscored/flattened column names, not dotted OTel names. Every trace filter except `trace_id`/`span_id` returns HTTP 400, and SpanName/ServiceName/ParentSpanID never populate. Verified the corrected mapping works (filters return counts, fields parse). Details inline.

**Note on portability:** the *core* span fields (`trace_id`, `span_id`, `operation_name`, `service_name`, `duration`, `span_kind`, `status_code`, `span_status`, `reference_parent_span_id`) are OpenObserve's built-in trace schema and are the same on every deployment — fixing those is a portable win. The *attribute-derived* fields (`workload_namespace`, `http_status_code`, `resource`, …) are flattened resource/span attributes whose names/presence depend on the instrumentation, so no hardcoded string is universally correct for them — schema discovery would be the durable fix (same story as `body` vs `message` for logs).

reckless-sherixx:
@mangglesh sorry for the delayed response. I have addressed all your comments

mangglesh (approved):

PR #619, opened 2026-07-17
fix(llm): make log query builder respect account context and real Signoz fields

mayankpande88:
# Description

Log searches through the AI agent could loop for minutes and come back with "no logs found" even when the logs existed. A real customer conversation showed ~30 SigNoz queries in a row, all empty, because the queries were built with field names SigNoz does not have (`pod`, `namespace`) instead of its real ones (`k8s.pod.name`, `service.name`, ...).

Before → now:

- **Account context now reaches the log query builder.** The operator-curated Global Context (deployment facts like real log field names and query conventions) was injected into the planner but never into the LLM call that builds the actual log query. It is now appended to that call's system prompt (size-capped, account-stable so prompt caching is preserved). Verified live: adding a marker instruction to an account's Global Context changes the generated query accordingly.
- **Correct SigNoz fallback fields.** When the backend's label discovery fails, the builder used to advertise `_body`/`namespace`/`pod` for every provider. For SigNoz those match nothing and every query silently returns zero rows. SigNoz now falls back to its OTel attribute names (`body`, `service.name`, `k8s.cluster.name`, `k8s.namespace.name`, `k8s.pod.name`, `trace_id`). Other providers are unchanged.
- **Resource search gets the same account context.** `resource_search` is the other custom-planner agent that translates user vocabulary into backend queries (K8s/Cloud/Datadog lookups); it now attaches the Global Context to its tool-call generation the same way. Audit note: metrics, PromQL, and all traces agents are ReAct-planner agents and already receive the Global Context through the planner's `<global_preferences>` block — no change needed there (a stale comment claiming that block was event-analysis-only is corrected in this PR). Web search, unified search, visualization, and GitHub log agents were reviewed and intentionally left untouched: they don't build telemetry-backend queries, and injecting operator context there only widens the blast radius of a badly worded instruction.
- **Query parsing tolerates LLM prose.** The query-generator LLM sometimes emits the JSON followed by extra prose ("Wait, let me reconsider..."), which failed the strict parse with `Unmarshal: there are bytes left`. Both the log and trace query builders now extract the JSON robustly (same helper already used by the kubectl intent path).

## Type of change

# How Has This Been Tested?

- [x] Unit tests replaying the exact production failure payloads (fenced JSON + trailing prose, two fenced blocks) against the log and trace query builders
- [x] Unit tests pinning the SigNoz fallback field list and the account-context propagation contract (in system message, byte-stable across calls, truncated when oversized, absent leaves prompt untouched)
- [x] Live end-to-end against a deployed stack (account with agent-source SigNoz + Loki): `@fetch_logs` built the query with real OTel field names and returned logs from SigNoz; inserting a Global Context instruction for the account visibly changed the generated query
- [x] `golangci-lint` clean; full `agents`, `tools/core`, `common` test suites pass

### DIFF
--- a/llm/llm-server/agents/agent_log_fetch.go
+++ b/llm/llm-server/agents/agent_log_fetch.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"nudgebee/llm/agents/core"
 	"nudgebee/llm/common"
+	"nudgebee/llm/config"
 	"nudgebee/llm/security"
 	"nudgebee/llm/services_server"
 	"nudgebee/llm/tools"
@@ -644,7 +645,46 @@ func truncateForLog(s string, n int) string {
 // (OriginalQuery, ConversationContext, the per-step query) lives in a single
 // human message so the upstream provider's prompt cache can hit on the static
 // system prefix across calls and conversations.
+// defaultCustomAgentAccountPromptBytes is the fallback cap for the account
+// GlobalContext fragment attached to custom-planner LLM calls when the
+// llm_server_agent_account_prompt_max_bytes config is unset/invalid.
+const defaultCustomAgentAccountPromptBytes = 8192
+
+// customAgentAccountPromptCap returns the byte cap for the account
+// GlobalContext fragment attached to custom-planner LLM calls (log/trace/
+// kubectl intent generators, resource search), so a large curated context
+// can't bloat every call. Configurable via
+// llm_server_agent_account_prompt_max_bytes (default 8192). ReAct agents
+// don't need this — the planner renders the GlobalContext into its human
+// message already. The cap makes no assumption about the document's layout —
+// content beyond it is simply dropped from the end.
+func customAgentAccountPromptCap() int {
+	if v := config.Config.LlmServerAgentAccountPromptMaxBytes; v > 0 {
+		return v
+	}
+	return defaultCustomAgentAccountPromptBytes
+}
+
 func buildLogIntentMessages(systemPrompt string, request core.NBAgentRequest) []llms.MessageContent {
+	// Propagate the account-wide GlobalContext (operator-curated deployment
+	// facts: real backend field names, id formats, naming conventions) into the
+	// query-generator call. Without this the generator only ever sees the
+	// discovered/fallback field list — when discovery fails, account-curated
+	// field guidance is the only correct signal available.
+	//
+	// Placed in the SYSTEM message deliberately: it is account-stable, not
+	// per-call (the system prompt is already account-scoped via the discovered
+	// fields list), so the cacheable prefix guarded by
+	// TestBuildLogIntentMessages_SystemPromptIsStable is preserved — only
+	// per-call inputs (Query/OriginalQuery/ConversationContext) must stay out
+	// of the system message.
+	if ap := strings.TrimSpace(request.AccountPrompt); ap != "" {
+		ap = core.TruncateHead(ap, customAgentAccountPromptCap())
+		systemPrompt += "\n\n**Account preferences (operator-curated for THIS deployment):**\n" +
+			"The notes below may state this backend's real log field names, id formats, and query conventions. " +
+			"When they name log fields for this backend, treat those names as part of the available Fields list.\n" +
+			ap + "\n"
+	}
 	var human strings.Builder
 	hasHints := false
 	if orig := strings.TrimSpace(request.OriginalQuery); orig != "" && orig != strings.TrimSpace(request.Query) {
@@ -728,6 +768,29 @@ tail SHOULD be 10000 so the grep has a meaningful window to scan.
 	return intent, nil
 }
 
+// defaultProviderLogFields is the field list advertised to the query-generator
+// LLM when the backend's label discovery (QueryLabels) fails or returns empty.
+// The where-clause keys are passed to the backend verbatim (no canonical
+// mapping), so the fallback must use each backend's REAL field names: Signoz
+// stores logs under OTel attribute names — the generic `_body`/`namespace`/
+// `pod` set matches no Signoz attribute and every query silently returns zero
+// rows.
+//
+// Signoz is special-cased because it has ONE fixed schema (OTel), so a
+// universally-correct fallback exists. Do NOT add hardcoded fallbacks for
+// schema-less backends like Elasticsearch: their field names depend entirely
+// on the customer's log shipper (filebeat `message`/`kubernetes.*`, OTel
+// `body`/`k8s.*`, custom pipelines), and any guessed list silently returns
+// zero rows for the setups it doesn't match — the exact failure mode this
+// function exists to fix. Those backends must rely on label/index discovery
+// and the account GlobalContext (propagated by buildLogIntentMessages).
+func defaultProviderLogFields(provider string) []string {
+	if strings.EqualFold(provider, "signoz") {
+		return []string{"body", "service.name", "k8s.cluster.name", "k8s.namespace.name", "k8s.pod.name", "trace_id"}
+	}
+	return []string{"_body", "namespace", "pod"}
+}
+
 // generateLogQuery returns the JSON-where envelope logs_execute consumes.
 // defaultIndex is the backend's account-default index (from get_default_provider);
 // empty for backends with no index concept (e.g. Loki).
@@ -736,7 +799,7 @@ func generateLogQuery(ctx *security.RequestContext, request core.NBAgentRequest,
 
 	fieldsProvided := len(fields) > 0
 	if !fieldsProvided {
-		fields = []string{"_body", "namespace", "pod"}
+		fields = defaultProviderLogFields(provider)
 	}
 
 	var b strings.Builder
--- a/llm/llm-server/agents/agent_log_fetch_test.go
+++ b/llm/llm-server/agents/agent_log_fetch_test.go
@@ -659,3 +659,67 @@ func TestMakeFetchResponse_PreviewsLogsWhenFileRefPresent(t *testing.T) {
 		assert.Equal(t, small, env["logs"])
 	})
 }
+
+// TestBuildLogIntentMessages_AccountPromptPropagation pins the contract for
+// account GlobalContext propagation into the query-generator call:
+//   - present  → appended to the SYSTEM message (account-stable, cache-safe)
+//   - absent   → system message is exactly the raw prompt (no empty block)
+//   - stable   → same AccountPrompt across calls yields byte-identical system
+//     messages, preserving the provider prompt-cache prefix
+//   - oversized → truncated via TruncateMiddle so a huge curated context
+//     cannot bloat every intent call
+func TestBuildLogIntentMessages_AccountPromptPropagation(t *testing.T) {
+	const sysPrompt = "Static system prompt. Extract log retrieval parameters."
+	const gc = "Log fields for this deployment: service.name, body, k8s.pod.name."
+
+	textOf := func(m llms.MessageContent) string {
+		var b strings.Builder
+		for _, p := range m.Parts {
+			if tp, ok := p.(llms.TextContent); ok {
+				b.WriteString(tp.Text)
+			}
+		}
+		return b.String()
+	}
+
+	t.Run("account prompt lands in system message only", func(t *testing.T) {
+		msgs := buildLogIntentMessages(sysPrompt, core.NBAgentRequest{Query: "logs for svc", AccountPrompt: gc})
+		assert.Len(t, msgs, 2)
+		sys, human := textOf(msgs[0]), textOf(msgs[1])
+		assert.Contains(t, sys, gc)
+		assert.Contains(t, sys, "Account preferences")
+		assert.NotContains(t, human, gc)
+	})
+
+	t.Run("no account prompt leaves system prompt untouched", func(t *testing.T) {
+		msgs := buildLogIntentMessages(sysPrompt, core.NBAgentRequest{Query: "logs for svc"})
+		assert.Equal(t, sysPrompt, textOf(msgs[0]))
+	})
+
+	t.Run("same account prompt is byte-stable across calls", func(t *testing.T) {
+		a := buildLogIntentMessages(sysPrompt, core.NBAgentRequest{Query: "logs for svc A", AccountPrompt: gc})
+		b := buildLogIntentMessages(sysPrompt, core.NBAgentRequest{Query: "errors in svc B last 1h", OriginalQuery: "why is B failing?", AccountPrompt: gc})
+		assert.Equal(t, textOf(a[0]), textOf(b[0]), "system message must not vary with per-call inputs")
+	})
+
+	t.Run("oversized account prompt is truncated", func(t *testing.T) {
+		huge := strings.Repeat("x", defaultCustomAgentAccountPromptBytes+50_000)
+		msgs := buildLogIntentMessages(sysPrompt, core.NBAgentRequest{Query: "logs", AccountPrompt: huge})
+		sys := textOf(msgs[0])
+		assert.Less(t, len(sys), len(sysPrompt)+defaultCustomAgentAccountPromptBytes+1024)
+	})
+}
+
+// TestDefaultProviderLogFields pins the fallback field vocabulary advertised
+// to the query-generator when backend label discovery fails: Signoz stores
+// logs under OTel attribute names — the generic `_body`/`namespace`/`pod` set
+// matches no Signoz attribute and every query silently returns zero rows.
+func TestDefaultProviderLogFields(t *testing.T) {
+	signozFields := defaultProviderLogFields("signoz")
+	assert.ElementsMatch(t,
+		[]string{"body", "service.name", "k8s.cluster.name", "k8s.namespace.name", "k8s.pod.name", "trace_id"},
+		signozFields)
+	assert.Equal(t, signozFields, defaultProviderLogFields("SigNoz"), "provider match must be case-insensitive")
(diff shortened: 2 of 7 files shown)


llm/llm-server/agents/agent_log_fetch.go
```diff
@@ -728,6 +756,20 @@ tail SHOULD be 10000 so the grep has a meaningful window to scan.
...
+// mapping), so the fallback must use each backend's REAL field names: Signoz
+// stores logs under OTel attribute names — the generic `_body`/`namespace`/
+// `pod` set matches no Signoz attribute and every query silently returns zero
+// rows.
+func defaultProviderLogFields(provider string) []string {
+	if strings.EqualFold(provider, "signoz") {
+		return []string{"body", "service.name", "k8s.cluster.name", "k8s.namespace.name", "k8s.pod.name", "trace_id"}
+	}
+	return []string{"_body", "namespace", "pod"}
+}
```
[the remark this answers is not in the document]
mayankpande88: Intentionally not adding an ES fallback — the situation differs from Signoz in a way that matters. Signoz has one fixed schema (OTel attribute names), so a universally-correct fallback exists. Elasticsearch has no canonical schema: field names depend entirely on the customer's shipper (filebeat uses `message`/`kubernetes.namespace`, an OTel exporter uses `body`/`k8s.namespace.name`, custom fluentd pipelines use anything). Hardcoding `message`/`kubernetes.*.keyword` would be right for filebeat setups and silently return zero rows for every other setup — the exact failure mode this PR fixes for Signoz. ES accounts instead rely on label/index discovery (already advertised in the prompt) and the account GlobalContext propagation added in this PR, which lets operators state their deployment's real ES field names. Added a code comment documenting this so it doesn't come up again.

llm/llm-server/agents/agent_log_fetch_test.go
```diff
@@ -659,3 +659,67 @@ func TestMakeFetchResponse_PreviewsLogsWhenFileRefPresent(t *testing.T) {
...
+// matches no Signoz attribute and every query silently returns zero rows.
+func TestDefaultProviderLogFields(t *testing.T) {
+	signozFields := defaultProviderLogFields("signoz")
+	assert.ElementsMatch(t,
+		[]string{"body", "service.name", "k8s.cluster.name", "k8s.namespace.name", "k8s.pod.name", "trace_id"},
+		signozFields)
+	assert.Equal(t, signozFields, defaultProviderLogFields("SigNoz"), "provider match must be case-insensitive")
+	assert.Equal(t, []string{"_body", "namespace", "pod"}, defaultProviderLogFields("loki"))
+	assert.Equal(t, []string{"_body", "namespace", "pod"}, defaultProviderLogFields(""))
+}
```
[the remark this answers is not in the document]
mayankpande88: Skipping — follows from the thread above: there is no ES-specific fallback to pin. The existing test already covers the contract: signoz gets the OTel set (case-insensitive match), everything else gets the legacy default.

llm/llm-server/agents/agent_log_fetch.go
```diff
@@ -644,7 +644,32 @@ func truncateForLog(s string, n int) string {
...
+	// field guidance is the only correct signal available.
+	//
+	// Placed in the SYSTEM message deliberately: it is account-stable, not
+	// per-call (the system prompt is already account-scoped via the discovered
+	// fields list), so the cacheable prefix guarded by
+	// TestBuildLogIntentMessages_SystemPromptIsStable is preserved — only
+	// per-call inputs (Query/OriginalQuery/ConversationContext) must stay out
+	// of the system message.
+	if ap := strings.TrimSpace(request.AccountPrompt); ap != "" {
+		ap = core.TruncateHead(ap, maxLogIntentAccountPromptBytes)
```
hsundar-git: low

**Approach is right; consider reusing the existing injection helper instead of a third bespoke rendering.**

This is now the *third* place account context is rendered into a prompt, each with different wording/handling:
- planner → `renderGlobalPreferencesBlock` (`utils.go:439`) wraps it in `<global_preferences>` in the **human** message
- `agent_llm.go:83` → `"Account Instructions:
" + common.SanitizePromptInput(request.AccountPrompt)` in the **system** message
- this PR → `"**Account preferences...**" + ap` in the system message, **raw** (no `SanitizePromptInput`)

Two asks:
1. **Run it through `common.SanitizePromptInput`**, matching the only other system-prompt injection site (`agent_llm.go`). GlobalContext is operator-curated (low risk), so this is defense-in-depth/consistency rather than a security bug — but the precedent for system-prompt injection sanitizes.
2. Since a sub-LLM call like this genuinely *can't* inherit the planner's `<global_preferences>` block, threading `request.AccountPrompt` in is the correct mechanism — but factor the sanitize+cap+format into one small shared helper (e.g. `AppendAccountContext(systemPrompt, request.AccountPrompt)`) so this path, the trace path, and any future tool-internal LLM call stay consistent and get account context for free.

llm/llm-server/agents/agent_log_fetch_test.go
```diff
@@ -659,3 +659,67 @@ func TestMakeFetchResponse_PreviewsLogsWhenFileRefPresent(t *testing.T) {
...
 	})
 }
+
+// TestBuildLogIntentMessages_AccountPromptPropagation pins the contract for
+// account GlobalContext propagation into the query-generator call:
+//   - present  → appended to the SYSTEM message (account-stable, cache-safe)
+//   - absent   → system message is exactly the raw prompt (no empty block)
+//   - stable   → same AccountPrompt across calls yields byte-identical system
+//     messages, preserving the provider prompt-cache prefix
+//   - oversized → truncated via TruncateMiddle so a huge curated context
```
hsundar-git, line 669: low

Doc nit: the comment says `TruncateMiddle`, but the code uses `core.TruncateHead` (keeps the first `maxLogIntentAccountPromptBytes`, drops the tail — consistent with the const's "dropped from the end" note). Worth aligning the wording.

Also consider whether keeping the **head** is the right half: if an operator puts the critical field-name guidance at the *end* of a long curated context, `TruncateHead` would drop exactly that. `TruncateMiddle` (keep head+tail) may be the safer choice for curated docs.

llm/llm-server/agents/agent_log_fetch.go
```diff
@@ -728,6 +753,29 @@ tail SHOULD be 10000 so the grep has a meaningful window to scan.
...
+// universally-correct fallback exists. Do NOT add hardcoded fallbacks for
+// schema-less backends like Elasticsearch: their field names depend entirely
+// on the customer's log shipper (filebeat `message`/`kubernetes.*`, OTel
+// `body`/`k8s.*`, custom pipelines), and any guessed list silently returns
+// zero rows for the setups it doesn't match — the exact failure mode this
+// function exists to fix. Those backends must rely on label/index discovery
+// and the account GlobalContext (propagated by buildLogIntentMessages).
+func defaultProviderLogFields(provider string) []string {
+	if strings.EqualFold(provider, "signoz") {
+		return []string{"body", "service.name", "k8s.cluster.name", "k8s.namespace.name", "k8s.pod.name", "trace_id"}
```
hsundar-git, line 789: > ✅ **Looks Good**

Good call special-casing only SigNoz (single fixed OTel schema → a universally-correct fallback exists) and the explicit comment warning against adding guessed fallbacks for schema-less backends like Elasticsearch. That's exactly the right boundary — a guessed field list there would reintroduce the same silent zero-rows failure this fixes.

hsundar-git:
## Summary of Changes

Fixes AI log/trace search returning "no logs found" despite logs existing, via three independent fixes in the log-fetch path: (1) account **GlobalContext** now reaches the query-generator LLM call, (2) SigNoz gets correct OTel fallback field names when label discovery fails, and (3) the query parser tolerates prose/fences around the LLM's JSON.

### Review Summary

**Services affected:** llm-server (Go)
**Risk level:** Low — additive, well-tested, no schema/API change.

|---|---|---|
| Consistency / hardening | Account context injected **raw** into the system prompt; the only other system-prompt injection site (`agent_llm.go:83`) wraps it in `common.SanitizePromptInput`. GlobalContext is operator-curated (low risk) → defense-in-depth, not a bug. — `agent_log_fetch.go` | low |
| Architecture | Third bespoke rendering of account context; extract a shared `AppendAccountContext()` helper (sanitize+cap+format) so all sub-LLM calls stay consistent. | low |
| Docs | Test comment says `TruncateMiddle`; code uses `TruncateHead`. Also consider whether keeping the head (vs middle) is right for curated docs. — `agent_log_fetch_test.go` | low |

---

### On the reviewer's questions

**Q: Is appending account context to the query-generator prompt the only way, or is there a better way?**

The *mechanism* is correct and, for this call site, essentially unavoidable. The query generator (`buildLogIntentMessages`) is a **separate LLM call** with its own system prompt — it does **not** run through the ReAct3 planner, so it can't inherit the planner's `<global_preferences>` block. Threading `request.AccountPrompt` into it is the right fix.

The *better way* is not a different mechanism but **not hand-rolling a third variant**. Account context is already rendered three different ways:
- planner → `renderGlobalPreferencesBlock` → `<global_preferences>` (human message)
- `agent_llm.go` → `"Account Instructions:" + SanitizePromptInput(...)` (system message)
- this PR → `"Account preferences..." + ap` (system message, unsanitized)

Recommend one shared helper (`AppendAccountContext(systemPrompt, request.AccountPrompt)`) that sanitizes + caps + formats once. Then this becomes a one-liner and every future sub-LLM call is consistent.

**Q: How can all custom agents get this account-context loading?**

They largely already do — the **loading is centralized**:
- `conversation.go` calls `toolcore.LoadActiveGlobalContext(ctx, accountId)` **once per turn** and composes it into `NBAgentRequest.AccountPrompt` (via `mergeAccountPrompts`). No agent re-loads from the DB.
- The value is then available **everywhere**: on `request.AccountPrompt`, and on `toolContext.AccountPrompt` (`executor_planner.go:3306`).

Consumption is automatic for the two *main* agent-LLM paths:
- **Planner agents** → `<global_preferences>` block (all ReAct3 agents).
- **`agent_llm`-based agents** → "Account Instructions" in the system prompt.

The only gap is **tool-internal / sub-LLM calls that build their own prompt** (this log query generator; the trace query generator; any future "LLM-inside-a-tool"). For those, the pattern to standardize is: read `request.AccountPrompt` / `toolContext.AccountPrompt` and pass it through the shared helper above. So the answer isn't "make each agent load it" — it's "route every sub-LLM system prompt through one injector," and the value is already loaded and waiting on the request/tool context.

> Suggest verifying the **trace** query-generation path gets the same treatment — it received the JSON-extraction fix here but I didn't see account-context propagation added to it.

---
*Automated PR Review*

llm/llm-server/agents/agent_log_fetch.go
```diff
@@ -644,7 +645,46 @@ func truncateForLog(s string, n int) string {
...
+	// field guidance is the only correct signal available.
+	//
+	// Placed in the SYSTEM message deliberately: it is account-stable, not
+	// per-call (the system prompt is already account-scoped via the discovered
+	// fields list), so the cacheable prefix guarded by
+	// TestBuildLogIntentMessages_SystemPromptIsStable is preserved — only
+	// per-call inputs (Query/OriginalQuery/ConversationContext) must stay out
+	// of the system message.
+	if ap := strings.TrimSpace(request.AccountPrompt); ap != "" {
+		ap = core.TruncateHead(ap, customAgentAccountPromptCap())
```
hsundar-git, line 682: low

(Re-review after force-push.) The **cap** is now nicely centralized via `customAgentAccountPromptCap()`, but the **render** block itself is still copy-pasted between here and `agent_resource_search.go`, with subtly divergent wording (here: bold heading + "The notes below may state..." sentence; there: plain heading, no guidance sentence). And neither site runs the text through `common.SanitizePromptInput`, unlike the third injection site `agent_llm.go:83`.

Folding both into one helper fixes drift and the sanitization gap in one place:

```go
// in agents/core (or a shared agents helper)
func AppendAccountContext(systemPrompt, accountPrompt string) string {
	ap := strings.TrimSpace(accountPrompt)
	if ap == "" {
		return systemPrompt
	}
	ap = TruncateHead(common.SanitizePromptInput(ap), customAgentAccountPromptCap())
	return systemPrompt + "

**Account preferences (operator-curated for THIS deployment):**
" + ap + "
"
}
```
Low severity since GlobalContext is operator-curated, but it removes the wording drift and makes "every custom-planner call gets account context the same way" literally true.

llm/llm-server/agents/agent_resource_search.go
```diff
@@ -165,6 +165,16 @@ Strategy:
...
 	}
+	// Same custom-planner gap as SkillsContext above: this LLM call never sees
+	// the planner's <global_preferences> block, so the operator-curated account
+	// GlobalContext (naming conventions like "users say the base service name,
+	// deployments carry -internal/-rr variants") must be attached here for the
+	// generated search calls to use the deployment's real vocabulary.
+	if ap := strings.TrimSpace(request.AccountPrompt); ap != "" {
+		ap = core.TruncateHead(ap, customAgentAccountPromptCap())
+		messages = append(messages, llms.TextParts(llms.ChatMessageTypeSystem,
+			"Account preferences (operator-curated for THIS deployment):\n"+ap))
```
hsundar-git, line 176: low

This is the second copy of the account-context injection, and its wording differs from the `buildLogIntentMessages` version (no bold heading, missing the "treat those names as part of the available Fields list" guidance sentence that actually tells the model what to *do* with the context). Recommend the shared `AppendAccountContext` helper (see comment on `agent_log_fetch.go`) so both calls emit identical, sanitized text.

llm/llm-server/agents/agent_log_fetch_test.go
hsundar-git, line 669: low

Still open from the prior revision: the docstring says `TruncateMiddle` but the code uses `core.TruncateHead` (keeps the head, drops the tail). Align the comment — and reconsider whether keeping the head is right, since an operator may place the key field-name guidance at the end of a long curated context.

llm/llm-server/agents/agent_resource_search.go
```diff
@@ -217,6 +228,13 @@ Strategy:
 			if tc.Tool == "" && tc.ToolName != "" {
 				tc.Tool = tc.ToolName
 			}
+			// Function-calling-style output ({"name": ..., "arguments": ...})
+			// is what Gemini and OpenAI models emit by habit even when asked
+			// for tool/input keys; without this alias every call is dropped
+			// and the keyword fallback runs instead.
+			if tc.Tool == "" && tc.Name != "" {
+				tc.Tool = tc.Name
```
hsundar-git, line 236: > ✅ **Looks Good**

Correct and complete: the `name` alias pairs with the existing `Arguments` (`json:"arguments"`) field already mapped into `Input` in the priority block below, so a full function-calling-style `{"name":..., "arguments":...}` emission now resolves both the tool and its input instead of silently dropping to the keyword fallback.

hsundar-git:
## Re-review (force-pushed → `4e94638e`)

The PR was reworked since my last pass and directly answers the "how do all custom agents get account context" question in code. Good progress.

### Addressed since last review ✅
- **Extended to `resource_search`** — account GlobalContext is now attached to that custom-planner LLM call too, not just log-fetch. This is the concrete "all custom agents" fix.
- **Config-driven cap** — `customAgentAccountPromptCap()` + new `llm_server_agent_account_prompt_max_bytes` (default 8192), distinct from `llm_server_max_gc_bytes`. Clean.
- **Documented the architecture** — `planner_react_3.go` comments now explain *why* ReAct agents need no per-agent wiring (planner renders `<global_preferences>` in the human message) and why custom-planner agents (`fetch_logs`, `resource_search`) must attach it to their own calls. This is the right mental model and worth having in-tree.
- **Bonus fix** — `resource_search` now aliases function-calling `{"name":...,"arguments":...}` output (pairs with the already-handled `arguments`→`Input`), so Gemini/OpenAI-style emissions no longer silently fall through to the keyword fallback. Correct and complete.

### Still open (all Low)
| Finding | Note |
|---|---|
| Render block still duplicated | The **cap** was centralized but the **injection text** is copy-pasted across `agent_log_fetch.go` and `agent_resource_search.go`, with divergent wording — the resource_search copy drops the bold heading and the "treat those names as available Fields" guidance sentence. Fold into one `AppendAccountContext()` helper. |
| No `SanitizePromptInput` | Both sites append the account text raw; the third injection site (`agent_llm.go:83`) sanitizes. Low risk (operator-curated), but the helper above closes it in one place. |
| Test docstring | Says `TruncateMiddle`; code uses `TruncateHead`. Also worth reconsidering head-vs-middle for curated docs whose key guidance may sit at the end. |

### Answering the two questions (now that the code does)
- **Better way / only way?** The mechanism (attach `request.AccountPrompt` to each custom-planner call) is correct and unavoidable for calls that bypass the planner prompt. The remaining polish is purely DRY: one shared `AppendAccountContext(systemPrompt, request.AccountPrompt)` (sanitize+cap+format) instead of two near-identical inline blocks.
- **How do all custom agents get it?** Loading stays centralized (`LoadActiveGlobalContext` once per turn → `NBAgentRequest.AccountPrompt`, also on `toolContext.AccountPrompt`). ReAct agents inherit it via `<global_preferences>`; custom-planner agents attach it per-call. This PR now covers `fetch_logs` + `resource_search`; any *future* custom-planner agent should call the shared helper so it's covered by construction.

**Verdict:** functionally solid, recommend merge after the (optional) helper consolidation + the two doc/sanitize nits.

---
*Automated PR Review (re-review)*

RamanKharchee (approved):

PR #637, opened 2026-07-29
fix(ui): cloud summary TDZ fix

thisissinghji:
# Description

Port of nudgebee/nudgebee-enterprise#33337 (merged 2026-07-01).

The hash-navigation `useEffect` in `[CloudAccountDetails].jsx` listed `filterOptions` in its dependency array, but the `const filterOptions = useMemo(...)` declaration came **after** the effect in the component body. Since a deps array is evaluated eagerly at the `useEffect(...)` call site, the binding was still in its temporal dead zone → `ReferenceError: Cannot access 'filterOptions' before initialization` (minified as `'w'` in production), crashing the cloud account details page on load.

Fix: moved the effect below the `filterOptions` declaration — no logic change, pure reorder (24 insertions / 24 deletions).

## Type of change

# How Has This Been Tested?

- [x] Identical patch already merged and verified in the enterprise repo (nudgebee/nudgebee-enterprise#33337); applied here cleanly with minor line offsets.

### DIFF
--- a/app/src/pages/cloud-account/details/[CloudAccountDetails].jsx
+++ b/app/src/pages/cloud-account/details/[CloudAccountDetails].jsx
@@ -104,30 +104,6 @@ const CloudAccounts = () => {
   const [selectedFilter, setSelectedFilter] = useState(null);
   const [selectedSubTab, setSelectedSubTab] = useState(0);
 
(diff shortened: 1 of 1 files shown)


blue4209211:
chanegs already in place

PR #655, opened 2026-08-14
fix(ui): stop /api/auth/* 404s breaking login on cold dev boot

### LINKED ISSUE #394: 🐛 [BUG] - Authentication Endpoints Return 404 When Frontend and API Server Start Simultaneously
### Pre-flight checklist

- [x] I've tried reproducing on the latest `main` (where possible).

### Version / commit

main

### Loom Recording

A Loom video demonstrating the issue and the workaround/fix:

https://www.loom.com/share/0c662799e14b49abb09dac8806d2be87

### Description

When the frontend application and the API server are started at the same time, opening the application in the browser causes NextAuth authentication requests to fail. The frontend attempts to fetch the session before the authentication endpoints are available, resulting in repeated 404 responses and an authentication redirect loop.

### Root Cause

The frontend attempts to access the NextAuth session endpoint before the authentication backend is fully available. Since `/api/auth/session` returns an HTML 404 page instead of JSON, NextAuth fails while parsing the response and reports a `CLIENT_FETCH_ERROR`.

### Workaround / Fix

Use the following startup sequence:

1. Start the frontend server.
2. Open the application in the browser.
3. Start the Go/API server.
4. Refresh the page if necessary.

Following this order allows the application to initialize correctly and authentication works as expected.

### Expected behavior

The application should either:

* Wait until the authentication service is available before requesting the session, or
* Gracefully retry failed authentication requests instead of entering a redirect/error loop.

### Actual behavior

The application fails to establish a session and logs the following errors:

```text
GET /api/auth/session 404

[next-auth][error][CLIENT_FETCH_ERROR]
Unexpected token '<', "<!DOCTYPE "... is not valid JSON

POST /api/auth/_log 404

GET /api/auth/signin?error=SessionRequired&callbackUrl=...
404

POST /api/auth/_log 404
```

The browser enters a redirect cycle involving `/api/auth/signin`, and authentication does not complete successfully.

### Steps to Reproduce

1. Start the frontend server.
2. Start the API/Go server at approximately the same time.
3. Open the application in the browser immediately after both services start.

### Frequency

Always

### Reproduction URL

https://github.com/nudgebee/nudgebee

### Screenshots

_No response_

### Logs

```bash
GET /api/auth/session 404 in 102ms (next.js: 63ms, application-code: 39ms)
[browser] [next-auth][error][CLIENT_FETCH_ERROR] 
https://next-auth.js.org/errors#client_fetch_error Unexpected token '<', "<!DOCTYPE "... is not valid JSON {
  error: {
    message: Unexpected token '<', "<!DOCTYPE "... is not valid JSON,
    name: 'SyntaxError',
    stack: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
  },
  message: Unexpected token '<', "<!DOCTYPE "... is not valid JSON,
  url: '/api/auth/session'
}
    at new Promise (<anonymous>) (<anonymous>)
 POST /api/auth/_log 404 in 143ms (next.js: 111ms, application-code: 32ms)
 GET /api/auth/signin?error=SessionRequired&callbackUrl=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fauth%2Fsignin%3Ferror%3DSessionRequired%26callbackUrl%3Dhttp%253A%252F%252Flocalhost%253A3000%252Fapi%252Fauth%252Fsignin%253Ferror%253DSessionRequired%2526callbackUrl%253Dhttp%25253A%25252F%25252Flocalhost%25253A3000%25252Fapi%25252Fauth%25252Fsignin%25253Ferror%25253DSessionRequired%252526callbackUrl%25253Dhttp%2525253A%2525252F%2525252Flocalhost%2525253A3000%2525252Fapi%2525252Fauth%2525252Fsignin%2525253Ferror%2525253DSessionRequired%25252526callbackUrl%2525253Dhttp%252525253A%252525252F%252525252Flocalhost%252525253A3000%252525252Fapi%252525252Fauth%252525252Fsignin%252525253Ferror%252525253DSessionRequired%2525252526callbackUrl%252525253Dhttp%25252525253A%25252525252F%25252525252Flocalhost%25252525253A3000%25252525252Faccounts%25252525252Faccount-form%25252525253FcloudProvider%25252525253DDISCORD 404 in 87ms (next.js: 64ms, application-code: 22ms)
POST /api/auth/_log 404 in 142ms
```

### Browsers (UI only)

Chrome

### OS

_No response_

reckless-sherixx:
# Description

Sign-in is unusable on a cold `npm run dev` boot: every request to `/api/auth/*` returns the HTML `/404` page instead of hitting the `[...nextauth]` catch-all, so next-auth tries to `JSON.parse` HTML and fails with `CLIENT_FETCH_ERROR: Unexpected token '<', "<!DOCTYPE "...`.

Root cause is the Next 16.2.12 **Turbopack** dev server: requests arriving during the cold-boot compile window are not matched against pages-router **dynamic** routes and fall through to `/404`. Critically, the bad resolution **latches for the whole process lifetime** — restarting does not reliably clear it, only a filesystem change under `app/src/pages/` does. That is why it looked random and survived restarts. It is most likely when the app boots before services-server, since the failed license fetch widens the window.

Two changes:

**1. `npm run dev` now uses webpack** (`next dev -p 3000 --webpack`). Webpack matches the catch-all correctly — it logs `○ Compiling /api/auth/[...nextauth] ...` and serves 200. Measured **3/3 clean runs on webpack vs 2/4 failures on Turbopack** under the same race on the same machine. Turbopack stays one command away as `npm run dev:turbo` for anyone who wants the speed and is not touching auth. This affects **dev only** — `next build`, CI and production are untouched.

**2. `signin.tsx` / `signin_v2.tsx` no longer call `getProviders()` in `getServerSideProps`.** That helper issues an HTTP self-call to `/api/auth/providers` — the page asking its own server for a route that may not be servable yet — which is what produced the server-side `CLIENT_FETCH_ERROR`. `authOptions.providers` is built synchronously from env at module load, so reading it directly is both faster and correct regardless of boot order. The existing `providersFromAuthOptions()` fallback (and its comment) already anticipated this; this promotes it to the only path and drops the round-trip. `signinUrl`/`callbackUrl` are only ever produced, never consumed — only `id` and `name` are read by the component — so the shape is unchanged in practice.

Rationale is recorded in `docs/architecture-decisions.md` with a reconsider-if condition, and the caveat is documented in `app/CLAUDE.md`.

> **Note on #394:** that issue was closed with an incorrect root cause — it attributed the 404s to services-server not being up yet. It isn't: I had services-server down and `/api/auth/*` still returned 200 in 2 of 4 runs. Start order only *widens* the race window. #394 has been reopened with the corrected diagnosis and full evidence in [this comment](https://github.com/nudgebee/nudgebee/issues/394#issuecomment-5292827187).

Fixes #394

## Type of change

# How Has This Been Tested?

All verification done with **services-server (port 8000) deliberately down** — the exact failing scenario.

- [x] **Reproduced the bug first.** Fired 4× `/api/auth/providers` + `/api/auth/session` concurrently at a cold-starting Turbopack server: all returned **404** while `/signin` returned 200, matching the reported log line-for-line. Confirmed the 404s **persisted after boot completed**.
- [x] **Ruled out our code.** Sibling static routes (`token`, `revoke`, `signup`) return 405 (registered) while only the catch-all 404s; a throwaway `pages/api/_probe/[...p].ts` registers instantly; a probe dynamically importing `@lib/license`, `@lib/authHooks`, `@lib/UserService`, `ldapts`, `next-auth` and `./auth/[...nextauth]` returns `ok` for every module. The route module compiles and evaluates fine — it simply was not registered.
- [x] **Verified the fix end-to-end** via `npm run dev`: all `/api/auth/*` → **200**, `/signin` → **200** with providers rendered, and **zero** `CLIENT_FETCH_ERROR` in the server log.
- [x] **Verified the sign-in POST path**: `POST /api/auth/callback/credentials` with a real CSRF token returns **JSON**, not an HTML 404. (It returns 401 because services-server is down — the expected downstream failure, not the routing bug.)
- [x] **Repeatability**: webpack 3/3 clean runs; Turbopack 2/4 failures under the identical race.
- [x] `npm run lint2` (oxlint + `tsc --noEmit` + prettier) — **0 errors, all clean**.

## Review notes — risks & counterarguments

- **Webpack dev is slower than Turbopack on a codebase this size.** That is the real cost of this PR. Mitigated by keeping `dev:turbo`; happy to flip the default back if the team would rather trade reliability for build speed.
- **The underlying bug is in Next, not in us.** Next 16.3.1 is released and may fix it, but that is a larger, unverified framework bump across the monorepo and was deliberately left out of this PR. If someone verifies 16.3.x is clean, the reconsider-if in the decision log says to revert to Turbopack.
- **`dev:turbo` can still hit this.** Documented recovery: touch any file under `app/src/pages/` to force the rescan.
- **Not fixed here (out of scope):** when services-server is unreachable, the credentials provider surfaces `Cannot read properties of undefined (reading 'length')` from `getUserByUsername` instead of a clear "backend unavailable" message. Worth a follow-up issue.


RamanKharchee (approved):

reckless-sherixx:
recheck

PR #696, opened 2026-08-27
fix(ui): point the agent-update snackbar at the agent upgrade docs

RamanKharchee:
# Description

The agent-update snackbar's "Refer to **this document**" link is dead.

It pointed at `/help/docs/installation/agent/installation/` — a **root-relative** path, so the browser resolved it against the app's own origin (`https://<your-app-host>/help/docs/...`). Nothing serves `/help`: there's no such page under `app/src/pages/`, and no rewrite for it in `next.config.js`. Every user who hit "Update the Agent Version to X" and clicked through got a 404.

This routes it through `docsUrl()` from `@lib/externalUrls`, already imported in this file for the two `/docs/features/` links, and targets the **upgrade** guide rather than the first-time install page — the snackbar only fires when an already-installed agent is behind the latest version, so upgrade steps are what the reader actually needs. Self-hosted deployments that set `NEXT_PUBLIC_DOCS_URL` now get their own docs host instead of a dead link.

## Type of change

No tracking issue — spotted while reading this component; nothing matching in the tracker.

# How Has This Been Tested?

`npm run lint2` passes end to end (oxlint 0 errors, `tsc --noEmit` clean, check-tdz clean, `prettier --check` clean).

**To verify as a reviewer**, on a cluster whose agent is behind the latest version (that's what gates the snackbar — `agent_version_latest != cluster.agent.version` on a `K8s` cluster that isn't `NOT_CONNECTED` and whose snackbar you haven't dismissed):

1. Select that cluster; the snackbar appears in the header.
2. Click **this document**.
3. Expect `https://docs.nudgebee.com/docs/installation/agent/installation/upgrade/` in a new tab — verified 200. Before this change it was `https://<app-host>/help/docs/installation/agent/installation/`, a 404.

<details>
<summary>Why a module const instead of an inline docsUrl() call</summary>

The snackbar message is a long single-line template literal. An inline `${docsUrl('/docs/installation/agent/installation/upgrade/')}` pushes it past Prettier's print width, and Prettier then breaks the literal across five lines — the call expression is what gives it a break point, where the previous long string had none. Hoisting keeps the message one line and `prettier --check` clean. Same shape as `EVENT_PAYLOAD_DOCS_URL` in `TriggerConfigSidebar.tsx`.

Module scope is safe here: `docsUrl` reads `NEXT_PUBLIC_DOCS_URL`, which Next inlines at build time, so there's no import-order or runtime-env hazard in evaluating it once at module load.

The trailing slash is deliberate: `…/upgrade` 301s to `…/upgrade/`, and every other `docsUrl()` call site in the app uses the trailing-slash form, so this skips a redirect hop and matches convention.

Out of scope, but worth flagging: the same `{{doc_url}}/help/docs/...` pattern appears in runbook-action `doc_link` values seeded by migrations (`V364`, `V378`, `V435`, …). Those are server-side templated against a different base, so they may or may not be broken — different mechanism, and applied migrations can't be edited. Worth a separate look.
</details>

### DIFF
--- a/app/src/components/common/header/Header1.jsx
+++ b/app/src/components/common/header/Header1.jsx
@@ -73,6 +73,8 @@ import CustomDrawer from '@shared/CustomDrawer';
 import ProductUpdatesDrawerContent from '@shared/widgets/ProductUpdatesDrawerContent';
 import { useProductUpdates } from '@hooks/useProductUpdates';
 
+const AGENT_UPGRADE_DOCS_URL = docsUrl('/docs/installation/agent/installation/upgrade/');
+
 const Header1 = ({ showBorder = false }) => {
   const { data } = useSession({ required: true });
   const router = useRouter();
@@ -505,7 +507,7 @@ const Header1 = ({ showBorder = false }) => {
         let snackMessage = '';
         let disconnectedService = [];
         setSnackbarOpen(true);
-        snackMessage = `<span>Update the ${baseTitle} Agent Version to ${res.data?.nudgebee_list_versions.agent_version_latest}. Refer to <a href="/help/docs/installation/agent/installation/" target="_blank" rel="noopener noreferrer">this document</a> for instructions on how to update the agent.</span>`;
+        snackMessage = `<span>Update the ${baseTitle} Agent Version to ${res.data?.nudgebee_list_versions.agent_version_latest}. Refer to <a href="${AGENT_UPGRADE_DOCS_URL}" target="_blank" rel="noopener noreferrer">this document</a> for instructions on how to update the agent.</span>`;
         if (cluster.agent?.connection_status) {
           const connectionStatus = cluster.agent?.connection_status;
           if (!connectionStatus.relayConnection) {


mayankpande88 (approved):

RamanKharchee:
Thanks for the review. Both comments raise the same point — module-level `docsUrl()` evaluation hurting testability — so answering once here rather than twice inline.

Keeping it at module scope, for three reasons:

1. **This file's test already mocks the module wholesale.** [`Header1.test.jsx:222`](https://github.com/nudgebee/nudgebee/blob/main/app/__tests__/components/common/header/Header1.test.jsx#L222) does `jest.mock('@lib/externalUrls', …)` with a `docsUrl` stub. `jest.mock` is hoisted above imports, so the module-level call hits the stub, not the real `process.env` read. There is nothing to stub at a finer grain here.
2. **The value is a build-time constant in the app.** `docsUrl` reads `NEXT_PUBLIC_DOCS_URL`, which Next inlines at compile time — it cannot change between renders, so re-evaluating it inside the component on every effect run buys nothing.
3. **It matches existing convention.** `TriggerConfigSidebar.tsx` declares `EVENT_PAYLOAD_DOCS_URL` the same way.

Worth noting the suggested inline form (`href="${docsUrl('…')}"`) is what the module const is avoiding: the snackbar message is a single-line template literal past the 150-char print width, and an inline call gives Prettier a break point it doesn't have today, so `prettier --write` reflows the literal across five lines. A local `const` inside the effect would avoid that too — it's just the weaker of the two placements given points 2 and 3.

PR #698, opened 2026-08-29
fix(llm): make gpt-5.x call tools for short live-state questions on ReAct3

hsundar-git:
# Description

GPT-5 models (gpt-5.6-terra, gpt-5.6-sol) answer short questions about live state — "get me pod count in nudgebee namespace" — with a sentence *promising* to fetch the data ("I'll retrieve the current pod count") or saying they cannot. Nothing is queried and the user gets no number. Gemini answers the same question correctly.

Three prompt instructions cause it. Gemini improvises past them; GPT-5 follows them literally. It is a prompt defect that GPT-5 exposed, not a model limitation — the same models run long tool chains once the instructions are corrected.

1. `k8s/aws/gcp/azure_lean` carried *"if the user asked only for the command, return the command string and do NOT execute it"*. That fired on plain data requests. Removed — both model families still return a command when one is genuinely asked for.
2. The answer-without-tools rule in `react_3_base` was circular and sat under a heading that primed *short = no tools*. Rewritten to key on **where the answer lives**, not on how short the question is.
3. Each lean agent now carries a worked example of emitting an action for a short live-state question, using its own CLI tool. This was the single biggest lever in testing.

Also widens the critique gate: a top-level answer that ran **zero tools this turn** and reads as a refusal is re-planned rather than returned.

Ports nudgebee-enterprise#37180 (merged to `prod`).

## Type of change

# How Has This Been Tested?

`make validate` passes on this branch (lint + full test suite).

Behaviour verified against a live cluster on the enterprise build of this same change — **18 runs, 18 correct, 0 critique interventions**:

| config | models | live-state | command request | greeting |
|---|---|---|---|---|
| `TOOL_SCHEMA_VALIDATION=*` | terra, sol, gemini | 8/8 ✅ | returns command ✅ | no tools ✅ |
| `TOOL_SCHEMA_VALIDATION=think` (default) | terra, sol, gemini | 5/5 ✅ | returns command ✅ | — |

**To verify as a reviewer:** point an account at a GPT-5 config and ask `get me pod count in <namespace>`. Before: a sentence promising to fetch it. After: the number, with a `kubectl_execute` step in the trace. Then ask `give me the kubectl command to count pods in <namespace>` and confirm it still returns the command without running it.

- [x] `make validate` (lint + tests)
- [x] 18 live runs across 3 models and both schema-validation settings (on the enterprise build of this diff)
- [ ] CI tests pass

## Scope limits — please read before approving

- **Only the Kubernetes agent was exercised.** The `aws_lean` / `gcp_lean` / `azure_lean` examples have never been run by a model — `@<agent>` does not route through the API path used for testing, so every run landed on `k8s_orchestrator`. Their tool schemas were checked against each tool's actual definition (all four take `{"command": …}`), but that is static verification only.
- Untested providers: anthropic, bedrock, vertexai, huggingface/Qwen.
- `looksLikeCapabilityRefusal` is a literal English phrase list. It is fail-safe — a miss just means no critique fires, which is today's behaviour — but it will not generalise to a model that phrases refusals differently.
- Expect a **one-off prompt-cache miss** on deploy: the Google AI explicit cache is keyed on a content hash, so edited prompts re-cache once.

### DIFF
--- a/llm/llm-server/agents/core/executor_response_formatter_test.go
+++ b/llm/llm-server/agents/core/executor_response_formatter_test.go
@@ -325,7 +325,7 @@ func TestBasePromptTemplate_GatesRcaFrameworkOnQuery(t *testing.T) {
 	// ...but the GENERAL final-answer mechanics stay for BOTH (not inside the gate).
 	assert.Contains(t, inv, "single `<final_answer>` block")
 	assert.Contains(t, qry, "single `<final_answer>` block")
-	assert.Contains(t, qry, "GREETING & SIMPLE QUERIES")
+	assert.Contains(t, qry, "GREETINGS & ANSWERING WITHOUT TOOLS")
 
 	// No leftover template tokens on either branch (balanced {{if}}/{{end}}).
 	assert.NotContains(t, inv, "{{")
--- a/llm/llm-server/agents/core/planner_react_3.go
+++ b/llm/llm-server/agents/core/planner_react_3.go
@@ -1820,13 +1820,23 @@ func (o *NBReActPlanner3) Plan(
 			// slowing down sub-agents like kubectl, logs, etc.
 			topLevel := o.isTopLevelAgent()
 			isInvestigation := IsInvestigationRequestTask(o.request.Query)
-			critiqueAllowed := o.enableCritique || (config.Config.LlmServerReActCritiqueEnabled && topLevel && isInvestigation)
+			// A top-level answer that ran NO tools and then claims it could not get
+			// the data is a refusal, not an answer. Short live-state questions ("get
+			// me pod count in X") classify as Query, so the refusal ships. Critique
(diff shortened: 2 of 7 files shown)


hsundar-git:
Tightened the refusal-marker list after a high-priority review finding on the enterprise backport (nudgebee-enterprise#37184).

The bare substring `"was executed"` matched benign multi-turn answers — a model summarising earlier work ("the command was executed in the previous step") would trip the gate and fire a needless critique LLM call. Replaced with `"no command was executed"`, `"no tool was executed"`, `"no query was executed"`.

The risk is specific to the per-turn step count in this same diff: before it, only the first turn of a conversation could reach the gate, so a back-reference could never match. Counting per-turn is what opens that window.

`make validate` passes.

blue4209211 (approved):

PR #714, opened 2026-09-03
fix(compose): configure llm service networking

PrashantBtkl:
## What changed

- Override the LLM server container port to 8000, matching the 8005:8000 Compose mapping.
- Route LLM service API calls to the Compose services-server alias instead of a container-local localhost endpoint.

## Validation

- docker compose --profile full config --quiet
- go test ./workspace

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -296,6 +296,8 @@ services:
         required: false
     environment:
       <<: *app-common
+      PORT: "8000"
+      SERVICE_API_SERVER_URL: http://services-server:8000
       LLM_SERVER_DB_URL: *db-url
       LLM_SERVER_WORKSPACE_RUNTIME: docker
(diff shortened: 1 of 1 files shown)


blue4209211 (approved):

blue4209211:
The two overrides are appropriate and are needed when `llm/llm-server/.env` contains the host-oriented example values (`PORT=8005` and `SERVICE_API_SERVER_URL=http://127.0.0.1:8000`). Without these overrides, the Compose port mapping and container-to-container API routing can break; the Docker workspace also expects the LLM server at `llm-server:8000`.

One minor follow-up before merge: please update the comment near `x-app-common` which currently says `SERVICE_API_SERVER_URL` is intentionally left at its code default. This PR now sets it explicitly in the `llm-server` service, so that comment becomes stale.

Issue #717, opened 2026-09-04
[REQUEST] - Add CubeAPM as an observability integration


reckless-sherixx:
### Function
Integrations

### Who asked for this
Prospect — active deal

### Summary
NudgeBee cannot connect to CubeAPM. A team that already runs CubeAPM as their observability platform has no way to point NudgeBee at their logs, traces or metrics, and no way to get their CubeAPM alerts into the NudgeBee inbox — so every troubleshooting, investigation and alert-triage feature is unavailable to them unless they migrate to a platform we already support.

This adds CubeAPM as a first-class observability provider at parity with the existing OpenObserve and Splunk integrations: connect with a URL (and optional tokens), test the connection, then query CubeAPM logs, traces and metrics from NudgeBee, receive CubeAPM alerts as investigable events, and create CubeAPM alert rules from NudgeBee.

### What this unblocks
Recorded as a coverage bet rather than a named deal. CubeAPM is a cost-focused APM that teams adopt specifically to avoid per-host Datadog/New Relic pricing, and those teams are exactly the profile that will not migrate observability platforms to try NudgeBee. Supporting it widens the set of prospects that can be onboarded as-is, with no migration step in the way.

### How we will know it worked
An admin can add a CubeAPM integration in Admin → Integrations, pass the connection test, set it as the default log/trace/metrics provider, and then: see CubeAPM logs in the Logs tab with working filters and Log Groups; see traces in the Traces tab with a working waterfall; chart CubeAPM metrics with PromQL autocomplete; and receive a CubeAPM alert as an event in the inbox with its chart and sample log attached as evidence. The concrete pass/fail signal is a CubeAPM account completing all four of those flows without falling back to another provider.

### What we are not doing instead
This consumes the integration slot for this cycle, so the knowledge-graph service-map flow source and the LLM agent tools — both of which OpenObserve also still lacks (see #499, #500) — stay unbuilt for CubeAPM. Those are additive follow-ups: none of the four flows above depend on them.

### Basic Example
1. Admin → Integrations → Observability → CubeAPM.
2. Enter the CubeAPM URL (e.g. `http://cubeapm.observability.svc:3140`). Tokens are optional — CubeAPM's query port is unauthenticated by default.
3. Test Connection → succeeds.
4. Tick "Make CubeAPM default Log Provider" (and/or Trace / Metrics).
5. Logs, Traces and Metrics tabs now query CubeAPM, in both the visual builder and Code mode.
[...]

PR #750, opened 2026-09-16
fix(ui): stop an unparseable log timestamp from blanking the Logs page

mangglesh:
# Description

Expanding a log row could blank the entire Logs page instead of showing that row's fields, and it stayed blank until reload. It happened whenever the backend reported the row's time as a date string rather than a number, which Dynatrace and Loggly both do. The time is now read in whatever form the backend sent, and one it cannot read leaves that line blank instead of taking the page down.

Refs #717

## Type of change

# How Has This Been Tested?

1. Open **Infra → (a cluster) → Monitoring → Logs** on a Dynatrace or Loggly account.
2. Click any log row to expand it.
3. The panel opens and lists the row's fields, **timestamp** among them as a readable date. Before this change the page went blank.
4. Repeat on Loki or Elasticsearch — unchanged.

# Risks

- The dropdown now guesses a time's unit from its magnitude, as the log table's date column already does. A backend using an unusual unit could show a plausible but wrong date. Both readers share the one guess, so they would be wrong together rather than disagree.

<details>
<summary>Engineering detail — root cause, evidence, what's not in this PR</summary>

## Root cause

`KubernetesLogDetails` → `renderLabelRow` (`app/src/components/k8s/common/KubernetesTable.jsx`) rendered the canonical `timestamp` label with a hardcoded nanosecond divisor:

```jsx
{key === 'timestamp' ? <Datetime value={new Date(value / 1000000)} /> : …}
```

No unit detection, no guard. What fills that label is `normalizeOutputLogLabels` (`api-server/services/observability/service.go`), which inverts each source's `GetLabelMapping()` and copies `Labels[providerKey] → Labels[canonical]` when the canonical key is absent. `dynatrace_logs.go` maps `"timestamp": "timestamp"` and `loggly.go` the same — both date strings.

A date string divided by `1e6` is `NaN` → `Invalid Date` → `Datetime` builds its tooltip with `Intl.DateTimeFormat#formatToParts`, which throws `RangeError: Invalid time value` on a non-finite date. It throws *during render*, so React's error boundary unmounts the whole screen rather than that subtree.

Worth noting `Datetime`'s own `parseDateValue` already carries a nanosecond heuristic — but it is unreachable from this path, because the caller divides and wraps in `new Date(...)` first, so `parseDateValue` receives a `Date` and returns it on line one.

## The change

- `epochToMillis` — `formatDateTime`'s existing unit detection (seconds / millis / micros / nanos by magnitude), extracted so it has one definition. `formatDateTime` now calls it; thresholds are unchanged.
- `parseEpochTimestamp` — handles a number or a numeric string (Loki and the OTel backends send nanoseconds as a string to keep the low digits), returns `NaN` for anything that is not a bare epoch so callers can fall back.
- `renderTimestampValue` — epoch first, then `parseDateValue` for ISO strings, then raw monospace text. `parseDateValue` is exported for it.
- `Datetime` — renders `emptyValue` when the parsed date is invalid, instead of reaching the throwing tooltip.

## Evidence

- New `logTimestampRendering.test.jsx` covers the unit matrix, the non-epoch cases, an ISO render, and that an unparseable value does not throw.
- It is a regression test, not a restatement: with the fix reverted the suite fails, with it applied it passes.
- `oxlint` — 0. `prettier --check` — clean. `jest --findRelatedTests` on the three changed files — no new failures against baseline.
- `tsc --noEmit` reports 4 errors in this checkout, all `Cannot find module` (`@dnd-kit/sortable`, `@dnd-kit/utilities`, `cron-parser`, `cronstrue`) in files this PR does not touch — packages declared in `package.json` but absent from the local `node_modules`. CI installs cleanly.

## Not in this PR

The same defect exists in the enterprise repo, where it was originally found via a CubeAPM account; that side also received a provider-level fix so its backend never emits a date string into this label. This is the general fix and is not CubeAPM-specific — CubeAPM is not part of this repo.

</details>


mangglesh:
Mirror PR: the enterprise half of this change is nudgebee/nudgebee-enterprise#38520. The two diffs are identical except that this one also exports `parseDateValue`, which the enterprise copy of `Datetime.jsx` already exported.

PR #751, opened 2026-09-16
fix(observability): stop a CubeAPM log row's details from blanking the Logs page

mangglesh:
# Description

Expanding a log row on a CubeAPM account blanked the entire Logs page instead of showing that row's fields. CubeAPM sent the row's timestamp as a date string where the details panel expects a number, so it built an invalid date and crashed the page while drawing it. CubeAPM now sends the form the panel reads.

Refs #717

## Type of change

# How Has This Been Tested?

1. Open **Infra → (a cluster on a CubeAPM account) → Monitoring → Logs**.
2. Click any log row to expand it.
3. The panel opens and lists the row's fields, with **timestamp** as a readable date. Before this change the page went blank and stayed blank until reload.
4. Repeat on a non-CubeAPM account — unchanged.

# Risks

- Dynatrace and Loggly put a date string in the same slot and would blank the page the same way. This fixes CubeAPM at the source rather than hardening the shared date formatter, so that class stays open — revisit if it is reported for another backend.

<details>
<summary>Engineering detail — root cause, evidence, side effects</summary>

## Root cause

Not in the CubeAPM row conversion — in the shared canonical-label pass.

`normalizeOutputLogLabels` (`api-server/services/observability/service.go:1586`, called from `FetchLogs`) inverts each source's `GetLabelMapping()` and copies `Labels[providerKey] → Labels[canonical]` whenever the canonical key is **absent**. `cubeAPMLogLabelMapping` maps `timestamp → _time`, and CubeAPM's `_time` is ISO-8601 (`2026-09-16T06:18:01.630729233Z`), so that string landed in `Labels["timestamp"]`.

The log-row dropdown renders that label as epoch **nanoseconds** — `KubernetesLogDetails` → `renderLabelRow` in `app/src/components/k8s/common/KubernetesTable.jsx:1081`:

```jsx
{key === 'timestamp' ? <Datetime value={new Date(value / 1000000)} /> : …}
```

An ISO string divided by `1e6` is `NaN` → `Invalid Date` → `Datetime`'s tooltip calls `Intl.DateTimeFormat#formatToParts`, which throws `RangeError: Invalid time value` on a non-finite date. Because it throws **during render**, the error boundary unmounts the whole Logs screen rather than that one cell.

## The change

`cubeAPMRowToOutputLog` now sets `out.Labels["timestamp"] = parsed.UnixNano()` from `_time`. Setting it here pre-empts the alias, which only fills an absent key. `_time` stays in `Labels` as the provider's own filterable field, and `OutputLog.Timestamp` stays ISO-8601 — that is the cross-provider contract the log-table column already reads. An unparseable `_time` sets no label at all, rather than a bad one.

## Evidence

- Two new tests in `cubeapm_logs_test.go`, one asserting the value survives a real `normalizeOutputLogLabels` pass.
- Proved they are regression tests, not restatements — with the fix stashed:
  ```
  --- FAIL: TestCubeAPMRowToOutputLogTimestampLabelIsEpochNanos
      Labels[timestamp] = <nil>; want epoch nanos 1789539481630729233
      after normalizeOutputLogLabels: Labels[timestamp] = "2026-09-16T06:18:01.630729233Z"; want epoch nanos 1789539481630729233
  ```
  The second line is the bug verbatim.
- `go test -race ./observability/` — ok, 12.9s. `go vet` — 0. `gofmt -l` — clean. `go build ./...` — ok.
- `make validate` cannot complete locally: the pinned `golangci-lint` is built with Go 1.25 against a 1.26.4 target and refuses to load its config. Lint runs in CI.
- Swept for other readers of `Labels["timestamp"]` in Go and in `app/` — the dropdown is the only one.

## Side effects, accepted

- CubeAPM's `QueryLabels` builds the label list from sampled record keys, so `timestamp` now also appears in the log label picker and in the LLM tool's label list. It sits alongside `_time`, `_msg` and `_stream`, which were always listed, and resolves back to `_time` via `cubeAPMFieldsFor`.
- `int64` nanoseconds exceed float64's exact-integer range, so the browser's `JSON.parse` rounds by ~128ns. The label is displayed to the minute.

## Mirror PR

This mirrors nudgebee/nudgebee-enterprise#38517, which is already merged there. The two diffs are identical — same 22 added lines in the source, same 57 in the test.

## Not in this PR

The durable fix is in the frontend — parse that label by shape and guard `Datetime` against an Invalid Date, so no provider can take a screen down through the error boundary. That was written and then explicitly deferred in favour of fixing CubeAPM at the source; it is what the Risks bullet refers to.

</details>


mangglesh:
Mirror of nudgebee/nudgebee-enterprise#38517, which is merged there. Identical diff.

RamanKharchee (approved):

